PR #10, opened 2016-06-13
Added a check for the ms prefix before calling toUpperCase

dmiller9911:
msPrefix is lowercase in inline styles.  I added a quick check be calling the toUpperCase on the first character of the prefix.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -24,7 +24,9 @@ export const vendorPrefix = (function () {
         .join('')
         .match(/-(moz|webkit|ms)-/) || (styles.OLink === '' && ['', 'o'])
     )[1];
-    return pre[0].toUpperCase() + pre.substr(1);
+    // ms prefix is lowercase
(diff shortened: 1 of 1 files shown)


clauderic:
Thanks for this. Just wondering if there isn't a more elegant approach to handling this before I merge

dmiller9911:
It somewhat depends on what styles you are using and what browsers you intend to support.  The o and Moz prefixes are most likely not needed at all, and the ms prefix is only needed for IE9.  Webkit and unprefixed are likely the only two scenarios you will hit since you are using 3dTransforms.

If you want to still support the other prefixes you can use something like: 
```
const prefixes = {
  ms: 'ms',
  webkit: 'Webkit',
  o: 'o',
  moz: 'Moz'
};
export const venderPrefix = (function () {
  ...
  return prefixes[pre] ? prefixes[pre] : '';
});
```
That method would allow you to easily deprecate unsupported prefixes later as well.


Or you could use a library similar to this for all prefixes:  https://github.com/rofrischmann/inline-style-prefixer

PR #22, opened 2016-06-22
Add option to specify how to lock the helper to the container

### LINKED ISSUE #21: Add support for having the helper stay in the list's bounds
Currently only it's half is inside the list.

I can submit a PR if needed.

zaygraveyard:


clauderic:
Thanks for this. What do you think of providing a percentage value instead of a string? So inside might be 0%, outside might be 100% and in the middle (default) could be 50% of the element's height. As well, I suppose you could also provide a fixed number

clauderic:
Also, have you checked whether this affects auto-scrolling? I have a feeling that bit of code may need tweaking as well

zaygraveyard:
> have you checked whether this affects auto-scrolling? I have a feeling that bit of code may need tweaking as well

My first take on adding this feature was to change the `minTranslate` and `maxTranslate`, but then I saw that they were used in the auto-scrolling.
So I decided to leave the `minTranslate` and `maxTranslate`, and only update the `updatePosition()` method.
Normally this shouldn't effect the auto-scrolling, I tested a simple vertical list and it works as before.

> What do you think of providing a percentage value instead of a string?

Good idea, but in that case I don't think the `lockToContainerEdges` is the correct prop to use.
We can maybe add a `lockOffset` that takes a percentage or a px value.
Or a `lockOffsetStart` and `lockOffsetEnd` for specifying asymmetric offsets.

Any thoughts?

clauderic:
Ah, fantastic. And yeah, `lockOffset` sounds good. If you feel strongly about supporting asymetric offsets, I think it should just be an array of two params provided to `lockOffset`

zaygraveyard:
Great, I'm on it

zaygraveyard:
PR updated

src/SortableContainer/index.js
```diff
@@ -229,11 +285,17 @@ export default function SortableContainer(WrappedComponent, config = {withRef: f
...
-				if (translate[axis] < this.minTranslate) {
-					translate[axis] = this.minTranslate;
-				} else if (translate[axis] > this.maxTranslate) {
-					translate[axis] = this.maxTranslate;
-				}
+				const [minLockOffset, maxLockOffset] = this.getLockPixelOffsets();
+				const minOffset = this.dimension / 2 - minLockOffset;
+				const maxOffset = this.dimension / 2 - maxLockOffset;
+
+				translate[axis] = Math.max(
```
clauderic: Great work! Only comment would be that this part of the code is a little obscure, I'd have a hard time wrapping my head around it if I had to debug this

clauderic:
Would you mind updating the README.md accordingly? :)

src/SortableContainer/index.js
zaygraveyard: Do you prefer if replace it with a `limit(min, max, value)` or an `if .. else if`?

zaygraveyard:
> Would you mind updating the README.md accordingly?

Sure no problem ^^

src/SortableContainer/index.js
clauderic: I'm fine with either, I think both would be a lot more legible and easy to understand.

zaygraveyard:
PR updated

src/SortableContainer/index.js
```diff
@@ -229,11 +285,15 @@ export default function SortableContainer(WrappedComponent, config = {withRef: f
...
-				}
+				const [minLockOffset, maxLockOffset] = this.getLockPixelOffsets();
+				const minOffset = this.dimension / 2 - minLockOffset;
+				const maxOffset = this.dimension / 2 - maxLockOffset;
+
+				translate[axis] = limit(
+					this.minTranslate + minOffset,
+					this.maxTranslate - maxOffset,
+					translate[axis]
+				);
```
clauderic: 👍 Hell of a lot more legible

README.md
```diff
@@ -102,6 +102,22 @@ More code examples are available [here](https://github.com/clauderic/react-sorta
 | useWindowAsScrollContainer | Boolean  | `false` | If you want, you can set the `window` as the scrolling container                                                                                                                                                             |
 | hideSortableGhost          | Boolean  | `true`  | Whether to auto-hide the ghost element. By default, as a convenience, React Sortable List will automatically hide the element that is currently being sorted. Set this to false if you would like to apply your own styling. |
 | lockToContainerEdges       | Boolean  | `false` | You can lock movement of the sortable element to it's parent `SortableContainer`                                                                                                                                             |
+| lockOffset                 | `OffsetValue`\* \| [`OffsetValue`\*, `OffsetValue`\*] | `"50%"` | The offsets to use when locking the movement of the sortable element. **See below for more info**. *This property has not effect of `lockToContainerEdges` is `false`.*         |
+
```
clauderic: How about something along the lines of:
> When `lockToContainerEdges` is set to `true`, this controls the offset distance between the sortable helper and the top and bottom edges of it's parent `SortableContainer`. Percentage values are relative to the height of the item currently being sorted. If you wish to specify different behaviours for locking to the *top* of the container vs the *bottom*, you may also pass in an `array` (For example: `["0%", "100%"]`).

README.md
```diff
@@ -102,6 +102,22 @@ More code examples are available [here](https://github.com/clauderic/react-sorta
...
+A single value is equivalent to a pair that same value (`"50%"` <=> `["50%", "50%"]`).
+
+```
+ +-------------------------------+        !   Sortable List                 !
+ |      lockOffset[0] = "50%" ↕  |        !+-------------------------------+!
+=|= Item 1 ========================       !|                               |!
+!|                               |!       =|= Item 5 ========================
+!+-------------------------------+!        |      lockOffset[1] = "50%" ↕  |
+!   Sortable List                 !        +-------------------------------+
+```
```
clauderic: I would probably ditch this section, I don't really understand the drawings 😂

zaygraveyard: I wanted to put an image but I didn't find a way to do it without adding the image to this repo.
But yeah I'll remove this section :smile:

README.md
zaygraveyard: Much better, thank you.

zaygraveyard:
PR updated

README.md
clauderic: You can always create a `/.github/` folder

zaygraveyard: Ok, and what do you think of these?
lockoffsetend
lockoffsetstart

clauderic: Hmm I think I'm going to make some gifs for it later, I wouldn't worry about it for the time being. Besides, there's nothing like a demo to illustrate something like this, so I'll be updating the gh-pages examples

clauderic:
Thanks for your hard work on this man! :)

zaygraveyard:
Happy to help :smile:

PR #23, opened 2016-06-22
Bugfix: Support drag handles with children

brianreavis:
This PR walks parents to find the actual drag handle instance. Without this, sorting won't work if a handle has any children (`e.target.sortableHandle` isn't always the actual handle). To reproduce, use this as a drag handle:

```js
render() { return <span><span>::</span></span>; }
```

Great library by the way!


clauderic:
Derp. Thanks for this!

PR #24, opened 2016-06-23
Bugfix: Fixed index state not being updated

brianreavis:
It seems that currently an item's index updates only *if* the item unmounts/remounts after the dragging ends ([source](https://github.com/clauderic/react-sortable-hoc/blob/master/src/SortableElement/index.js#L21-L37))? I was getting wonky behavior after my first drag/drop because an element's `index` wasn't being updated at the end of dragging (even though I reordered the data outside of the control like in the examples).

Take this PR with a grain of salt. I couldn't totally wrap my head around where this library keeps it's "source of truth". It seems to be in a few places – I tried to cover the bases, but could be approaching this wrong.

Any ideas?

### DIFF
--- a/src/Manager.js
+++ b/src/Manager.js
@@ -1,5 +1,6 @@
 import find from 'lodash/find';
 import sortBy from 'lodash/sortBy';
+import {arrayMove} from './utils';
 
 export default class Manager {
 	refs = {};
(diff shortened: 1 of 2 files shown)


clauderic:
Hmmm. Yeah that's pretty silly on my part, not sure how I missed that. I'm not sure I'd handle it that way though, because this approach is a little naive in the sense that it assumes that the state will be updated, which may or may not be true. If there is no `onSortEnd` logic set up in the enhanced `SortableContainer` component to update the state, nothing will change when an item is dropped and all items will return to their initial positions.

I think it would probably be a lot easier to handle this using `componentWillReceiveProps` in `SortableElement` instead. Thoughts?

brianreavis:
Good call! Yea, I think that's a better and more flexible approach. I might not have time to tackle it in the next few days. If you want to take a stab at it that'd be awesome :) Otherwise I'll try to get to it when I have time

clauderic:
No worries, I'll take care of it 👍

clauderic:
Just to follow-up, I think I know why I didn't spot this earlier. I've been using ```key={`item-${index}`}```, which means the `SortableItem` indexes were never actually updating. If you use a key that doesn't rely on index, this no longer holds true, which I'm assuming is what happened in your case. I'll be issuing a fix shortly

brianreavis:
Just tried out your fix in `0.0.3` – works great! Thanks man

PR #25, opened 2016-06-23
Fix helperClass

ahmedelgabri:
when adding multiple classes with `classList.add` they must be like this `'foo', 'bar'` currently they are applied like this `'foo, bar'`. So this will fix this issue and will allow to add multiple classes. 

keep in mind that the space after the backtick is required.

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -129,7 +129,7 @@ export default function SortableContainer(WrappedComponent, config = {withRef: f
 				}
 
 				if (helperClass) {
-					this.helper.classList.add(helperClass.split(' '));
+					this.helper.classList.add(...(helperClass.split(' ')));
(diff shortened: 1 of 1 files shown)


src/SortableContainer/index.js
```diff
@@ -129,7 +129,7 @@ export default function SortableContainer(WrappedComponent, config = {withRef: f
 				}
 
 				if (helperClass) {
-					this.helper.classList.add(helperClass.split(' '));
```
zaygraveyard: Why not `this.helper.classList.add(...(helperClass.split(' ')));`?

ahmedelgabri: That's definitely better, didn't think about it. Updated the PR.

clauderic:
Thanks guys!

PR #27, opened 2016-06-24
Bugfix: Add support handling `margin`s on scrollable items. (fixes #26)

### LINKED ISSUE #26: BUG: `margin`s are not taken into account in the re-positioning of the items
### How to reproduce
1. Go to the [demo page](http://clauderic.github.io/react-sortable-hoc/)
2. Add `.Showcase__style__stylizedItem {margin-bottom: 10px;}` to the styles.
3. Drag the first item of the list bellow the second and don't release the mouse button.

### What was expected
The second item slides all the way to the top.

### What happens
The second item slides up but not all the way to the top (leaving a margin above it).

### Reason
The items are translated by `dimension` with is either `width` or `height`, [which in turn are `offsetWidth` and `offsetHeight` respectively](https://github.com/clauderic/react-sortable-hoc/blob/master/src/SortableContainer/index.js#L100)
But `offsetWidth` and `offsetHeight` do not include the margins [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetHeight).

### Snapshots
image
---
image

zaygraveyard:
Only support's [margin collapsing](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing) if all items have the same margins.
[How to uncollapse margin for adjacent siblings](http://stackoverflow.com/questions/1828804/how-do-i-uncollapse-a-margin#answer-1828940)
[How to disable margin collapsing](http://stackoverflow.com/questions/19718634/how-to-disable-margin-collapsing#answer-19719427)

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -1,7 +1,7 @@
 import React, {Component, PropTypes} from 'react';
 import ReactDOM from 'react-dom';
 import Manager from '../Manager';
-import {closest, events, vendorPrefix, limit} from '../utils';
+import {closest, events, vendorPrefix, limit, getElementMargin} from '../utils';
 import invariant from 'invariant';
 
 // Export Higher Order Sortable Container Component
@@ -100,13 +100,19 @@ export default function SortableContainer(WrappedComponent, config = {withRef: f
 				let {axis, onSortStart, helperClass, hideSortableGhost, useWindowAsScrollContainer} = this.props;
 				let {node, collection} = active;
 				let index = node.sortableInfo.index;
+				const margin = getElementMargin(node);
 
 				let containerBoundingRect = this.container.getBoundingClientRect();
 
 				this.node = node;
+				this.margin = margin;
 				this.width = node.offsetWidth;
 				this.height = node.offsetHeight;
 				this.dimension = (axis == 'x') ? this.width : this.height;
+				this.marginOffset = {
+					x: this.margin.left + this.margin.right,
+					y: Math.max(this.margin.top, this.margin.bottom)
+				};
 				this.boundingClientRect = node.getBoundingClientRect();
 				this.index = index;
 				this.newIndex = index;
@@ -118,8 +124,8 @@ export default function SortableContainer(WrappedComponent, config = {withRef: f
 
 				this.helper = this.document.body.appendChild(node.cloneNode(true));
 				this.helper.style.position = 'fixed';
-				this.helper.style.top = `${this.boundingClientRect.top}px`;
-				this.helper.style.left = `${this.boundingClientRect.left}px`;
+				this.helper.style.top = `${this.boundingClientRect.top - margin.top}px`;
+				this.helper.style.left = `${this.boundingClientRect.left - margin.left}px`;
 				this.helper.style.width = `${this.width}px`;
 
 				if (hideSortableGhost) {
@@ -339,12 +345,13 @@ export default function SortableContainer(WrappedComponent, config = {withRef: f
 				if (transitionDuration) {
 					node.style[`${vendorPrefix}TransitionDuration`] = `${transitionDuration}ms`;
 				}
+
 				if (index > this.index && (sortingOffset + offset >= edgeOffset)) {
-					translate = -this.dimension;
+					translate = -(this.dimension + this.marginOffset[axis]);
 					this.newIndex = index;
 				}
 				else if (index < this.index && (sortingOffset <= edgeOffset + offset)) {
-					translate = this.dimension;
+					translate = this.dimension + this.marginOffset[axis];
 
 					if (this.newIndex == null) {
 						this.newIndex = index;
--- a/src/utils.js
+++ b/src/utils.js
@@ -47,3 +47,21 @@ export function limit(min, max, value) {
     }
     return value;
 }
+
+function getCSSPixelValue(stringValue) {
+    if (stringValue.substr(-2) === 'px') {
+        return parseFloat(stringValue);
+    }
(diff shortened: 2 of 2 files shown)


src/SortableContainer/index.js
```diff
@@ -339,12 +341,21 @@ export default function SortableContainer(WrappedComponent, config = {withRef: f
...
 				}
+
+				const margin = getElementMargin(node);
+				// margins merge so we subtract them
+				const delta = (
+					axis === 'x'
+					? this.width  + this.margin.left + this.margin.right  - margin.left
+					: this.height + this.margin.top  + this.margin.bottom - margin.top
+				);
+
```
zaygraveyard: @clauderic 
I found a problem with this de-merging of margins, when `margin.left > this.margin.left` it doesn't work.
There are other cases where all the items below need re-positioning.
So please don't merge yet

zaygraveyard: @clauderic 
I found that margin collapsing is a very hard to solve problem and a rarely used feature.
So I didn't add support for it.
This PR is now ready for merging.

clauderic: Hmm. I'm not sure it's that uncommon. Have you looked at how other libraries handle this? Also, have you tried with negative margins 🤔

zaygraveyard: > I'm not sure it's that uncommon.

I don't know, but I personally never used it and can't think of a use case for it in a list.

> Have you looked at how other libraries handle this?

All the other reordering libraries that I found change the order of the element in the DOM, so don't need to handle this case.
One technique used to animate changes between two states of the DOM that stood-out was the [FLIP animation](https://aerotwist.com/blog/flip-your-animations/#the-general-approach) used by [react-flip-move](https://github.com/joshwcomeau/react-flip-move)

> Also, have you tried with negative margins 🤔

No, I can't say that I have. I'll try it and let know

zaygraveyard: Negative margins work as expected :smile: 
The only supported feature is margin collapsing which happens on vertical margins only (top and bottom) and when they are of the same type (positive or negative)

zaygraveyard:
@clauderic any updates on this PR?

src/SortableContainer/index.js
```diff
@@ -339,12 +346,13 @@ export default function SortableContainer(WrappedComponent, config = {withRef: f
 				if (transitionDuration) {
 					node.style[`${vendorPrefix}TransitionDuration`] = `${transitionDuration}ms`;
 				}
+
 				if (index > this.index && (sortingOffset + offset >= edgeOffset)) {
-					translate = -this.dimension;
+					translate = -this.dimensionWithMargins;
```
clauderic: I could be wrong, but I believe margin collapsing can easily be solved if you change this line to something along the lines of:
```js
(axis === 'y') ? -(this.dimension + this.margin.top) : -(this.dimension + this.margin.left)
```

And if you change https://github.com/clauderic/react-sortable-hoc/pull/27/commits/90c5a73c228774b22b547ecb84860eafd2de2bd7#diff-683d0c001c717e14072ccfa0e6202f1fR355 to:

```js
(axis === 'y') ? this.dimension + this.margin.bottom : this.dimension + this.margin.right
```

zaygraveyard: Sorry for the delay, I think I missed the notification message for some reason :confounded:.
First-off margin collapsing only happens vertically (`margin-top` and `margin-bottom`), so we should add the horizontal margins (`margin-left` and `margin-right`).
```js
this.marginOffset = {
    x: this.margin.left + this.margin.right,
    y: Math.max(this.margin.top, this.margin.bottom)
}
```

As for using the max of the two margins, it only works if we assume all items have the same margins.
I had done a similar logic but after seeing all the possible cases I decided it's easier and cleaner to say we don't support margin collapsing.
That said I'm open to supporting this case (all item have the same margins).

clauderic: Are you sure margin collapsing wouldn't happen for a horizontal sorting set-up?

I think I'd be okay with only supporting this case

zaygraveyard: @clauderic 
From [MDN's Mastering margin collapsing](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing)
> Top and bottom margins of blocks are sometimes combined (collapsed) into a single margin whose size is the largest of the margins combined into it, a behavior known as margin collapsing.

[Live example](http://jsfiddle.net/fhdjznwe/)

And I updated this PR to support this case.

clauderic:
Thanks for updating your PR 👍 Let me test your latest fix out and if all looks good I'll merge ASAP :)

src/utils.js
```diff
@@ -47,3 +47,21 @@ export function limit(min, max, value) {
     }
     return value;
 }
+
+function getCSSPixelValue(stringValue) {
+    if (stringValue.substr(-2) === 'px') {
+        return parseFloat(stringValue);
+    }
+    return 0;
+}
```
clauderic: Not sure I understand why this function is necessary? Why not just use parseInt / parseFloat directly?

zaygraveyard: It's just a safe garde to ensure that `stringValue` is a pixels value since as mentioned on [StackOverfow](http://stackoverflow.com/questions/14275304/how-to-get-margin-value-of-a-div-in-original-javascript#answer-14275343):
> Fair warning: What you get back may not be in pixels. For instance, if I run the above on a p element in IE9, I get back "1em".

If you find that such browsers are not supported, I can remove this function and `parseFloat` directly.

PR #28, opened 2016-06-24
🤗  Support nested SortableComponents.

Cap32:


clauderic:
Hmm. What's the use-case? Doesn't using the `collection` prop already solve this? 

Regardless, I think using `stopPropagation` might conflict with other components, no?

Cap32:
👍🏿 Oh. The `collection` prop works. I didn't realize this prop on this issue. Thanks.

PR #33, opened 2016-07-16
add ref to manager when disabled prop goes from true to false

wmdmark:
This fixes an issue for when a `SortableElement` goes from a disabled to a non-disabled state.

### DIFF
--- a/src/SortableElement/index.js
+++ b/src/SortableElement/index.js
@@ -28,7 +28,7 @@ export default function SortableElement (WrappedComponent, config = {withRef: fa
             addOrRemove(collection, this.ref)
         }
         componentDidMount() {
-            this.addOrRemoveRef(this.props.disabled)
+            this.addOrRemoveRef(!this.props.disabled)
         }
         componentWillReceiveProps(nextProps) {
             const {index} = this.props;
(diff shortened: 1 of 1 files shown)


clauderic:
Ah, good catch. Thanks for submitting this! Looks good, though I believe you missed the case for when the prop goes from `enabled` to `disabled` though no?

wmdmark:
@clauderic oops, did not think to handle that since it was the other way that was broken for me. Does [this](https://github.com/clauderic/react-sortable-hoc/pull/33/commits/448c501b669723eeb90f81d00d0c0a4653f68dd5) look OK for handling both cases? Have not had a chance to test yet.

wmdmark:
@clauderic any thoughts on this? would be nice to get this merged so I can stop using my fork.

clauderic:
Hey @wmdmark, sorry I didn't get back to you. There's another PR to the same effect (https://github.com/clauderic/react-sortable-hoc/pull/34), I'm undecided as to which implementation I prefer, but I will merge one of the two as soon as I get the chance

clauderic:
I just merged #34, going to close this out. Thank you for submitting this.

PR #37, opened 2016-07-27
Added support for a custom container (via getContainer prop)

bvaughn:
I've also added a new Storybook section demonstrating how this property can be used to integrate with more complex 3rd party components such as `FlexTable`.

This PR also updates the version of 'react-virtualized' used in 'devDependencies' to 7.15 (so that a FlexTable example could be added using the new getContainer prop) and adds 'react' to 'devDependencies' since it was missing.


bvaughn:
Hey @clauderic :)

I tried to keep the change set size minimal and stick with something close to what we discussed earlier via Gitter. Happy to make changes or adjustments!

src/SortableContainer/index.js
```diff
@@ -58,9 +59,9 @@ export default function SortableContainer(WrappedComponent, config = {withRef: f
 			};
 		}
 		componentDidMount() {
-			let {contentWindow} = this.props;
+			let {contentWindow, getContainer} = this.props;
 
-			this.container = ReactDOM.findDOMNode(this);
+			this.container = (typeof getContainer == 'function') ? getContainer(this.getWrappedInstance()) : ReactDOM.findDOMNode(this);
```
bvaughn: I was on the fence about whether this should return a React element or a DOM node. Thoughts?

clauderic: I think I'd lean towards DOM node, simply because it would enable the use of elements outside of React's scope, for instance, `document.body`

bvaughn: That's what I thought also. (So that's what it currently does.)

joaoreynolds:
I'm trying to use this new feature (getContainer) and I'm having trouble making it work. Can you provide an example? I seem to be returning the wrong object in this function. I've tried sending the ref object, as well as the dom object returned by `findDOMNode`

bvaughn:
Hey @joaoreynolds 

You can see an example here:
https://github.com/clauderic/react-sortable-hoc/blob/master/src/.stories/index.js#L97-L143

PR #43, opened 2016-08-30
Helper Z-Index

jetpackjarrett:
This will prevent the dragged item from falling behind items with high z-indexes (like modals).

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -125,6 +125,7 @@ export default function SortableContainer(WrappedComponent, config = {withRef: f
 
 				this.helper = this.document.body.appendChild(node.cloneNode(true));
 				this.helper.style.position = 'fixed';
+				this.helper.style['z-index'] = 999999;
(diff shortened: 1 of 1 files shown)


clauderic:
Hey @jetpackjarrett, sorry for not getting back to you sooner. 

The reason why I didn't implement something like this is I think z-indexes should be something that's managed by the end-user of the library, not by the library itself. Using z-index 999999 is generally a code-smell. As well, other libraries may already be assigning something to z-index 999999 (or greater), in which case the user would still need to overwrite the value you've assigned here with !important (also a code-smell).

If anything, it may be worth mentioning in the docs that this is something that can be achieved through applying a z-index to the `helperClass`.

jetpackjarrett:
@clauderic thanks for the feedback. For whatever reason, the helperClass would not apply a z-index. I'll give it another go.

clauderic:
Hmm, odd. Perhaps an issue with stacking contexts? (though unlikely, since the helper gets appended to the body)

PR #54, opened 2016-09-15
Fix lockOffset Documentation

HashDot:
Fix escape typo


clauderic:
Thanks for catching this!

PR #59, opened 2016-09-23
fix server side rendering with ReactDOMServer

binary1101:
ReactDOMServer provides window, but not document, so it still breaks server rendering.


clauderic:
Looks great, thanks for submitting this! :shipit:

binary1101:
can you release it to npm?

clauderic:
Yep, just published `0.0.9` to npm 👍

PR #66, opened 2016-10-05
support changable collection

krisl:
my collection name changes ( I keep the path to the parent in the collection name, eg 'grandparent:0:parent:5:children' )  and the parents themselves are sortable, as such their indexes can change.
I didn't want to bind a new closure for each `onSortEnd` because I have so many items.
Anyhow, one PR


clauderic:
Thanks for submitting this, just merged back into master and released it with `0.1.0` 👍

PR #69, opened 2016-10-13
Update arrayMove util.

canvaskisa:
Update arrayMove util to return the copy of array, instead of mutating the passed one.

I suppose, it is a bad practice to mutate an existing array with this util, it can cause some unpredictable side-effects. It works fine for local react component state, but can cause errors, when using it outside of the local state (flux stores or redux store, for example) by mutating the passed array.


clauderic:
Looks good, clean and simple 👍

PR #71, opened 2016-10-24
Fix timing issue in Chrome

chrisvasz:
This change fixes a hard-to-debug and reproduce issue in Chrome. In one window, I can't sort anything at all, either in my application or even on the GitHub pages site. In another Incognito window in the same version of Chrome, I can sort just fine. I think it is a timing issue related to the way `setTimeout` and the `setState` inside `handlePress` interact with each other. I tried eliminating the `setTimeout` when `pressDelay === 0` and the (alleged) timing issue went away.

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -111,7 +111,11 @@ export default function SortableContainer(WrappedComponent, config = {withRef: f
 				this.manager.active = {index, collection};
 
 				if (!distance) {
-					this.pressTimer = setTimeout(() => this.handlePress(e), this.props.pressDelay);
+					if (this.props.pressDelay === 0) {
+						this.handlePress(e);
+					} else {
+						this.pressTimer = setTimeout(() => this.handlePress(e), this.props.pressDelay);
+					}
 				}
 			}
 		};


clauderic:
Hmmm 🤔 Interesting. And are you seeing similar issues when there is a `pressDelay`?

clauderic:
If you're not seeing the issue in incognito, might I suggest you take a look at your extensions? You may have a chrome extension enabled that's causing the issues you're seeing (and then the extension is disabled in incognito, which is the chrome default for most extensions)

chrisvasz:
I tried disabling all my extensions and it made no difference. That matches my empirical experience -- seems like it works on some days and not others. The most recent environmental change was an update to Chrome 55 beta, but the fact that it works in Incognito mode rules out Chrome version.

I also tried adding a `pressDelay` of 200, and neither this proposed fix nor the existing code worked for me in that case. Haven't explored any further.

clauderic:
Hmm. I'm unable to replicate this at the moment. Let me update to the latest version of chrome and I'll get back to you.

frontendphil:
+1 on this. When `pressDelay` is `0` then there is no reason to break the sync chain of calls here.

PR #73, opened 2016-10-25
add touch cancel support for utils

### LINKED ISSUE #72: onTouchCancel not accounted for - leaves cloned element painted on DOM
@clauderic - First off amazing lib we are currently using it in our production application with amazing results. Our use case is we are building a react application then using phoneGap to build to IOS and Android. I noticed an issue where if the user is currently dragging an element and enters split-screen mode on Ipad or hits the menuButton on IOS ```onTouchCancel``` gets fired.

Because this event is not accounted for in the utils / events list ```onHandleEnd``` never gets called and thus the clonedElement remains painted on the DOM. Off the top of my head I would think adding onTouchCancel to the event.end object would solve this but wanted to get your thoughts before opening a PR.

Take a look at this video: http://recordit.co/a51WvhIQVP

Daadler6:
treat touchCancel event like touchEnd to un-mount cloned element from DOM. This should resolve #72

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -13,7 +13,7 @@ export function arrayMove (arr, previousIndex, newIndex) {
 export const events = {
 	start: ['touchstart', 'mousedown'],
 	move: ['touchmove', 'mousemove'],
-	end: ['touchend', 'mouseup']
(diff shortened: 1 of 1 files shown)


clauderic:
This is perfect, thanks!

PR #75, opened 2016-10-26
Various code improvements:

ralf-strehle:
- better usage of let and const
- unified indentation 2 spaces
- blank line between methods
- unfolding of if { ... }
- some small linting errors fixed

### DIFF
--- a/src/Manager.js
+++ b/src/Manager.js
@@ -2,26 +2,31 @@ import find from 'lodash/find';
 import sortBy from 'lodash/sortBy';
 
 export default class Manager {
-	refs = {};
-	add(collection, ref) {
-		if (!this.refs[collection]) this.refs[collection] = [];
+  refs = {};
 
-		this.refs[collection].push(ref)
-	}
-	remove(collection, ref) {
-		let index = this.getIndex(collection, ref);
+  add(collection, ref) {
+    if (!this.refs[collection]) {
+      this.refs[collection] = [];
+    }
+    this.refs[collection].push(ref);
+  }
 
-		if (index !== -1) {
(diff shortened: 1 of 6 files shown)


ralf-strehle:
Dear Clauderic,

we are using your fine react sortable in an e-learning CMS react application and I decided to transpile the source files ourselves with babel because some time ago we had strange stability problems with the dragging not starting on some machines. I did some basic source formatting to get a better understanding of the code, and maybe these small improvements can be useful for you.

Kind regards,
Ralf

clauderic:
Hey @ralf-strehle, thanks for submitting this.

I'd be open to accepting your changes for `let` vs `const` usage, so if you want to create another PR for those changes, I'd be happy to discuss and review there.

However, regarding the "linting errors" you mention, it all depends on what linting rules you follow. There is no clear consensus, so it's really all a matter of personal preference.

This project follows the [xo](https://github.com/sindresorhus/xo) code style, please spend some time familiarizing yourself with it if you wish to contribute back to this repo :)

ralf-strehle:
Hi Clauderic,

thank you for your reply. I can try what you suggest but would need some time.

But I think that there is more about it than just let, const and var. E.g. most linters agree that it is not good practice to mix tab and spaces. Indenting should be unified. If you do not like 2 spaces (I like it because of all those callbacks in js) then it is good practice to stick to 4 spaces. Also === is better than ==. Also, when semicolon is used at the end of a statement, then I think that usage should be consistent. 

So I am surprised that XO does not catch these code formatting issues. When I look at https://github.com/sindresorhus/xo then I see many options, maybe the linter needs some configuration.

Maybe spend some second thoughts on it :)

PR #78, opened 2016-10-27
Some refactoring

ralf-strehle:
- better usage of let and const
- blank lines between class methods and properties
- do not switch indentation between tabs and spaces in same code file
- consistent usage of semicolon at end of statement
- replace super() with super(props) in constructor
- initialize local component state in constructor
- replace == with type safe ===
- initialize direction with 0
- avoid confusion with declaration of index in SortableElement's componentWillReceiveProps
- some other minor code improvements

### DIFF
--- a/src/Manager.js
+++ b/src/Manager.js
@@ -2,31 +2,31 @@ import find from 'lodash/find';
 import sortBy from 'lodash/sortBy';
 
 export default class Manager {
-  refs = {};
+	refs = {};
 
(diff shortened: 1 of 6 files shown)


ralf-strehle:
Hi Clauderic,

I hope that it is easier to diff and merge now, and that you can agree with the improvements.

Kind regards,
Ralf

clauderic:
Looks good, thanks!

PR #84, opened 2016-11-10
Add a getHelperDimensions fn to control SortableItem size. Fixes #83

### LINKED ISSUE #83: Changing height of SortableItem after drag start breaks size calculations
We wanted to shrink the dragged element `onSortStart`, but at that time the height of the element is already read and subsequently used for move coordinate calculations. Is there any thing we can hook into in order to set height _before_ it is read, or do we need to start working on an enhancement to *react-sortable-hoc*?
If element 1 is 500px high and we set it to 50px on sort start, it will displace the following elements 450px too far towards the top of the container.

nervetattoo:
Pass in a helper to dictate item dimensions:
```jsx
<SortableList getHelperDimensions={({node}) => ({width: node.offsetWidth, height: 50})} />
```

### DIFF
--- a/src/.stories/Storybook.scss
+++ b/src/.stories/Storybook.scss
@@ -102,6 +102,9 @@
     }
 }
 
+.shrinkedHelper {
+    height: 20px !important;
+}
 
 :global {
     body {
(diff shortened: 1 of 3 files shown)


clauderic:
Looks great! Thanks again @nervetattoo!

PR #94, opened 2016-11-21
The proper way to clear interval

Neurostep:
Just found a place where used `clearTimeout` instead of `clearInterval` which could cause some side-effects.


clauderic:
Ha! Would you look at that 🙈 

Thanks man!

Issue #98, opened 2016-11-29
HOCs should consume props they use


frontendphil:
The HOCs should consume the props they use internally and not hand them down to the wrapped components. Currently I need to sort out all internal props so that they do not end up as props on the elements I'm rendering.

clauderic:
Yeah, this has been discussed already in #95 and should be resolved soon. If you feel like getting your hands dirty, all PRs are welcome!

clauderic:
This was addressed in #100, thank you!

derwaldgeist:
Hm, I don't think that consuming the `index` was a good idea. Knowing the index of the element in the list is important for a lot of use-cases.

PR #100, opened 2016-11-30
clauderic/react-sortable-hoc#98 do not pass private props to wrapped …

frontendphil:
I did a couple of things in this PR.

- all HOCs are also exported in their lowercase version, i.e. `sortableElement`. This is good style as they are used as functions and not components and all functions which are not used along with `new` should be lowercase
- I removed `getHelperDimensions` as even though it was used, the retrieved values were overwritten anyways.
- I removed the `halfHeight` and `halfWidth` instance properties. Those can be easily computed each time they are needed. Also this reduces the amount of maintenance as you don't need to keep `height` and `width` in sync with their `half` relatives.

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -43,7 +43,11 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
 				}
 			},
 			lockToContainerEdges: false,
-			lockOffset: '50%'
+			lockOffset: '50%',
+			getHelperDimensions: ({node}) => ({		
+ -				width: node.offsetWidth,		
+ -				height: node.offsetHeight		
+ -			})
 		};
 
 		static propTypes = {
@@ -67,7 +71,8 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
 				PropTypes.string,
 				PropTypes.arrayOf(PropTypes.oneOfType([PropTypes.number, PropTypes.string]))
 			]),
-			getContainer: PropTypes.func
+			getContainer: PropTypes.func,
+			getHelperDimensions: PropTypes.func
 		};
 
 		static childContextTypes = {
@@ -163,17 +168,18 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
 			const active = this.manager.getActive();
 
 			if (active) {
-				const {axis, onSortStart, helperClass, hideSortableGhost, useWindowAsScrollContainer} = this.props;
+				const {axis, getHelperDimensions, helperClass, hideSortableGhost, onSortStart, useWindowAsScrollContainer} = this.props;
 				let {node, collection} = active;
 				const {index} = node.sortableInfo;
 				const margin = getElementMargin(node);
 
 				const containerBoundingRect = this.container.getBoundingClientRect();
+				const dimensions = getHelperDimensions({index, node, collection});
 
 				this.node = node;
 				this.margin = margin;
-				this.width = node.offsetWidth;
-				this.height = node.offsetHeight;
+				this.width = dimensions.width;	
+ -				this.height = dimensions.height;
 				this.marginOffset = {
 					x: this.margin.left + this.margin.right,
 					y: Math.max(this.margin.top, this.margin.bottom)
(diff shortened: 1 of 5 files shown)


clauderic:
Hey @frontendphil, thanks for submitting this!

Unfortunately, in it's current state, I can't accept your PR completely. Would you mind spliting your changes into individual PRs?

I'd be happy to accept both your changes for omitting passing down private props as well as your changes for halfHeight/halfWidth.

However, I don't think changing the HOCs to lowercase would be a good idea at this stage, as this would be a breaking change.

As for `getHelperDimensions`, it was introduced in #84 by request, I'll need to investigate further.

richmeij:
Looks like the HOCs are exported as lowercase in addition to the existing uppercase variants, so that wont be a breaking change, would it?

richmeij:
Im a bit biased on the `halfWidth` / `halfHeight` though. I know they seem trivial, but In my case it made it easier to reason about the calculations (less parentheses, clear use). Plus, the whole point of doing the calculations only once, was to not have to do them again later on, so how can that be more 'amount of maintenance'? :)

Looking at the PR, the single use of the `halfWidth` and `halfHeight` consts in `animateNodes()` IS rather pointless indeed, but Im not so sure about `this.halfWidth` and `this.halfHeight`.

src/SortableContainer/index.js
```diff
@@ -1,11 +1,13 @@
 import React, {Component, PropTypes} from 'react';
 import ReactDOM from 'react-dom';
-import Manager from '../Manager';
-import {closest, events, vendorPrefix, limit, getElementMargin} from '../utils';
+import { omit } from 'lodash'
```
richmeij: `import omit from 'lodash/omit'` would be better

frontendphil:
Alright, yeah since I export both in uppercase and lowercase this change is not breaking. 

And regarding the `halfHeight` discussion. I would argue that reasoning of what goes on in this component is actually a problem, that needs tackling in the future. But it's not in these simple formulas but in the long and nested `if` statements. 

However having `halfHeight` as an instance property directly connects `height` and `halfHeight` with each other. This means everyone who wants to change the `height` property needs to have the knowledge about `halfHeight` and make sure to keep them in sync. I think that the gain of removing this internal coupling outweighs the disadvantages.

clauderic:
@frontendphil hadn't noticed you were exporting both at first, my bad ❤️ 

I also tend to agree with you when it comes to halfWidth / halfHeight, and it also feels cleaner, even if it results in duplicated computations.

richmeij:
Allright, sounds good. :+1:

frontendphil:
Since everyone seems to be ok with these changes, could we merge them then?

richmeij:
I think the only thing still open is the removal of `getHelperDimensions`. Havent looked in to this myself yet.

src/SortableContainer/index.js
```diff
@@ -168,22 +163,17 @@ export default function SortableContainer(WrappedComponent, config = {withRef: f
...
 
 				const containerBoundingRect = this.container.getBoundingClientRect();
-				const dimensions = getHelperDimensions({index, node, collection})
 
 				this.node = node;
-				this.width = dimensions.width
-				this.height = dimensions.height
 				this.margin = margin;
 				this.width = node.offsetWidth;
 				this.height = node.offsetHeight;
```
frontendphil: If you look closely here, you can see that the values retrieved from `getHelperDimensions` are set to `this.height` and `this.width` and are overwritten only one line later with `node.offsetWidth` and `node.offsetHeight`. This means that `getHelperDimensions` was never used even though the method existed. 

Most probably this is not desirable but should be resolved in another PR. In this PR however I removed the double assignment, to avoid confusion.

clauderic:
Let's not overcomplicate things too much. I restored the getHelperDimensions method and removed the duplicate this.height / this.width declaration that slipped by the first time around.

Let's get this merged 🚢 😄

davidgoli:
This is a pretty unfortunate implementation, as it makes hard-coded assumptions about what props wrapped components do and do not consume. Would it be possible to check the omitted list against eg. the propTypes of the passed in component, or otherwise offer a whitelist to the hoc wrapper?

PR #106, opened 2016-12-08
fix: workaround for mozilla bug

funnel-mark:
Firefox has a fun bug, described here:

https://bugzilla.mozilla.org/show_bug.cgi?id=548397
window.getComputedStyle() returns null inside an iframe with display: none

This causes react-sortable-hoc to throw `TypeError: can't convert null to object`

The workaround is to return an array with a single fake mozilla-prefixed style in it if window.getComputedStyle() returns falsy.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -18,7 +18,11 @@ export const events = {
 
 export const vendorPrefix = (function () {
     if (typeof window === 'undefined' || typeof document === 'undefined') return ''; // server environment
-    let styles = window.getComputedStyle(document.documentElement, '');
+    // fix for:
+    //    https://bugzilla.mozilla.org/show_bug.cgi?id=548397
+    //    window.getComputedStyle() returns null inside an iframe with display: none
+    // in this case return an array with a fake mozilla style in it.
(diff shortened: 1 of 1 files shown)


clauderic:
Hey @funnel-mark, thanks for flagging this and providing a detailed PR!

Based on your proposed implementation, I wonder if there aren't other situations where window.getComputedStyle() might return a falsy value?

funnel-mark:
As far as I can understand from the documentation on MDN `window.getComputedStyle()` should either return a CSSStyleDeclaration or throw an error. It's only because of this bug that it returns null, it should never happen otherwise.

clauderic:
I see. Thanks for the clarification! 👍

mikhail-polontech:
Since Firefox version 62, window.getComputedStyle() returns a CSSStyleDeclaration object with length 0 (instead of null) inside an iframe with display: none.
See this comment: https://bugzilla.mozilla.org/show_bug.cgi?id=548397#c53.

CSSStyleDeclaration object with length 0 is not a falsy value, so the fix above does not work anymore.

PR #115, opened 2016-12-31
perf(lodash): use vanilla omit

### LINKED ISSUE #114: Replace `_.omit` with a vanilla method (perf, size, future proofing)
Reference: https://github.com/Semantic-Org/Semantic-UI-React/issues/864

### Speed
In some recent benchmarks over at Semantic-UI-React, the `_.omit` method has proven to be _extremely_ slow.  I achieved **>12,000 times performance gain with a 4 line vanilla replacement**.

### Size
Even with cherry-picking, the `_.omit` method requires many dependencies, totaling 96 modules, 3,172 lines, and ~84kb (unminified).  This will be bundled by users consuming `react-sortable-hoc` whether or not they are using lodash.

### It will be dropped in v5

[Per the author](https://github.com/Semantic-Org/Semantic-UI-React/issues/864#issuecomment-261152847), `_.omit` will be dropped in v5 due to some design concerns.

levithomason:
Fixes #114 

### Limitations
The lodash method accepts keys to omit as many arguments, `_.omit(obj, 'a', 'b', 'c')` or as an array `_.omit(obj, ['a', 'b', 'c'])`.  I've only supported the former as this project only implemented that usage.

### Testing
I noticed there is no test rig.  I added the below snippet to `utils.js` and ran it with `babel-node` to test the util.  I also tested the storybook by hand and all seems well.

```js
const obj = { foo: 'foo', bar: 'bar', baz: 'baz' }

// No omitted keys
console.log(omit(obj))
//=> { foo: 'foo', bar: 'bar', baz: 'baz' }

// Some omitted keys
console.log(omit(obj, 'foo', 'bar'))
//=> { baz: 'baz' }

// All omitted keys
console.log(omit(obj, 'foo', 'bar', 'baz'))
//=> {}

// Omitted non-matching keys
console.log(omit(obj, 'not in there'))
//=> { foo: 'foo', bar: 'bar', baz: 'baz' }
```

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -1,10 +1,9 @@
 import React, {Component, PropTypes} from 'react';
 import ReactDOM from 'react-dom';
-import omit from 'lodash/omit'
 import invariant from 'invariant';
 
 import Manager from '../Manager';
-import {closest, events, vendorPrefix, limit, getElementMargin, provideDisplayName} from '../utils';
+import {closest, events, vendorPrefix, limit, getElementMargin, provideDisplayName, omit} from '../utils';
 
 // Export Higher Order Sortable Container Component
 export default function sortableContainer(WrappedComponent, config = {withRef: false}) {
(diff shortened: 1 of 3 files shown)


clauderic:
Looks good! Thanks for putting this together @levithomason 

LGTM 🚢

Issue #122, opened 2017-01-19
Select in SortableItem doesn't have value if is draggable


tomasztomys:
If you drag an item with select, the value of select set to value of first item's select and return to itself value after drop. But text on the left of select is correct.

![image](https://cloud.githubusercontent.com/assets/7750691/22104907/c05cc772-de41-11e6-93ab-c98eb5d79bc2.png)

https://jsfiddle.net/60848oue/

clauderic:
Hey @tomasztomys, this was fixed by #123 and shipped in `0.4.8` 🎉  Thanks for reporting this!

PR #123, opened 2017-01-19
Fix inputs values.

tomasztomys:
#122

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -214,11 +214,12 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
 				};
 
 				const fields = node.querySelectorAll('input, textarea, select');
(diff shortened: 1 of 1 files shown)


clauderic:
Hey @tomasztomys, thanks for submitting this 😄 

I slightly altered your implementation of the pre-population logic to account for cases where the select/textarea/input wouldn't have a `name` attribute. Instead, I'm simply running the same `querySelectorAll` query which returns the elements in the same order as the first query, and therefore makes it easy to simply map the values from one to the other via their index.

🚢 LGTM

tomasztomys:
Are you sure that all nodes in cloned node will have the same order in each situation? Find by name is safer

clauderic:
AFAIK They will always be returned in document order, which will not change when the element is cloned. Find by name would only be reliable assuming the inputs were named and unique, no?

Issue #136, opened 2017-02-03
Buttons not supported for shouldCancelStart


joelworsham:
It would be nice if the tag `button` was also in the default list. Of course, I can add my own function to modify this, but it just seems like `<button>`s would be expected for this behavior.

https://github.com/clauderic/react-sortable-hoc/blob/master/src/SortableContainer/index.js#L40

clauderic:
Hey @joelworsham, that's a good observation, thanks for pointing that out. If you have time, is this something you'd like to submit a PR for? I'd be happy to review if so 😄

Robinfr:
Made PR for you since I like this library. :smile:

Robinfr:
Can be closed since the PR was merged.

joelworsham:
Thanks guys!

PR #142, opened 2017-02-10
Button will now be canceled in SortableContainer

Robinfr:
Added the buttuon to `shouldCancelStart` to allow buttons to be clicked within SortableContainers.

See issue #136


clauderic:
Thank you! 😄

PR #145, opened 2017-02-13
Added story for distance option

Robinfr:
Added new story to storybook for the distance option.


clauderic:
Thanks, great idea! 👍

PR #147, opened 2017-02-14
Implemented dragging prop in SortableElement

Robinfr:
I've updated SortableElement to inject the `isDragging` prop into the WrappedComponent. I've also changed the SortableContainer to allow the SortableElement to re-render before making a clone by making a clone after the state change has been performed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,13 +5,13 @@ import {SortableContainer, SortableElement, arrayMove} from './src/index';
 import range from 'lodash/range';
 import random from 'lodash/random';
 
-const SortableItem = SortableElement(({height, value}) => (
+const SortableItem = SortableElement(({height, value, isDragging}) => (
     <div style={{
         position: 'relative',
         width: '100%',
         display: 'block',
         padding: 20,
-        backgroundColor: '#FFF',
+        backgroundColor: isDragging ? 'orange' : '#FFF',
         borderBottom: '1px solid #EFEFEF',
         boxSizing: 'border-box',
         WebkitUserSelect: 'none',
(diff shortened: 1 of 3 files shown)


clauderic:
Hmm, hadn't considered that approach. Have you tested alongside the `collection` prop?

I actually wonder if we couldn't try using `ReactDOM.unstable_renderSubtreeIntoContainer` instead of having to call `setState` and then clone the node, which still feels a little hacky

Robinfr:
@clauderic no I haven't tried that yet. I'll do that tomorrow.

The callback of `setState` is meant for this kind of stuff so even though I understand it feels a bit hacky, I think it's perfectly fine. Using something that has the word `unstable` in it on the other hand feels a bit dangerous..

clauderic:
Hmm. It's unstable in the sense that the API may change in future releases, but it is still the preferred way of rendering a component into a different container.

Robinfr:
Do you want me to try to implement the `renderSubtreeIntoContainer` in a different branch or you want to see if it can be included with this?

Robinfr:
I've updated the PR to also take into account the `collection` prop.

Robinfr:
Resolved conflicts.

Robinfr:
@clauderic do you have any intention to merge this? Otherwise I will close this..

tizzle:
Hey @clauderic,

any news on this?

Ricardo-Marques:
+1 on this. I have also wanted to keep track of whether an element was being dragged and soon discovered that the element is cloned before it re renders. This feels like an elegant solution.

minikomi:
Can this PR get another look? It seems useful for large components.

alidavut:
Changing some styles by dragging would be very helpful. Can you update the PR?

Chaoste:
@clauderic Hey, is there any chance that this can be merged soon? We would like to give focus-like styles to the dragged component which appears to be impossible at the moment.

PR #153, opened 2017-02-28
Bug/offset parent

kaoDev:
At the moment it can happen that the offset of a draggged node is calculated incorrectly. With the introduced utility the offset is calculated relative to the next positioned parent.


clauderic:
Hey @kaoDev, 

Thanks for submitting this, and apologies for the delay getting back to you. Do you have a jsfiddle I could look at to better understand the scenario you're trying to solve with this PR?

Also, there are merge conflicts, would you mind rebasing?

kaoDev:
hey,
here is a fiddle showing the bug https://jsfiddle.net/s7rag1z4/1/
It happens as soon as the Sortable Element is wrapped inside another node like in this example the div around the SortableElement.

kaoDev:
oh damn, the autoformatter did way too much after the merge

src/SortableContainer/index.js
```diff
@@ -430,7 +445,7 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
           top: offset.top + node.offsetTop,
           left: offset.left + node.offsetLeft,
         };
-        if (node.parentNode !== this.container) {
+        if (getOffsetParent(node, this.container) !== this.container) {
```
kaoDev: relevant change in index.js
instead of comparing the parent node with the container, use the offsetParent (first positioned node)

PR #160, opened 2017-03-10
Default to `pointer-events: none` on Sortable helper

clauderic:
Because... trackpad ninjas.

https://twitter.com/Vjeux/status/840264657424146436

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -288,6 +288,7 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
         this.helper.style.width = `${this.width}px`;
         this.helper.style.height = `${this.height}px`;
         this.helper.style.boxSizing = 'border-box';
+        this.helper.style.pointerEvents = 'none';
 
         if (hideSortableGhost) {
           this.sortableGhost = node;


src/SortableContainer/index.js
```diff
@@ -288,6 +288,7 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
         this.helper.style.width = `${this.width}px`;
         this.helper.style.height = `${this.height}px`;
         this.helper.style.boxSizing = 'border-box';
+	this.helper.style.pointerEvents = 'none';
```
bvaughn: White space police! 🚓   Tab spotted!👮

clauderic: 😂

The one time I'm lazy and decide to use the built-in github editor 🙄

clauderic: Never will I ever again.

bvaughn: GOTCHA! 🥇

rauchg: Prettier-bot would not have allowed this 🤖

clauderic: (╯°□°）╯︵ ┻━┻

bvaughn: It may be time to step away from open source, Clauderic

ahmedisam99:
Why!! just why??!!?
don't you think you should use a property to allow users to specify whether to disable pointer events or not? @clauderic 
It's causing a trouble with all users wanting to specify a custom cursor while dragging the list-item.

josias-r:
@clauderic Took me a while until I finally found why the cursor is not behaving as it is in the examples. Ahmedisam99 pointed it out already, in a rude manner though, that this change breaks the cursor styles. Would be nice if this was configurable. Also it's misleading that the Examples are not up-to-date where this change has not yet been made.

Issue #170, opened 2017-03-26
React Sortable HOC interfering with React Draggable


szabolcsx:
Hi,

Based on the [following](https://embed.plnkr.co/N4iYHYE9gPLp4Y3NHbhk/) code I tried to implement sortable table columns with React Sortable HOC and React Virtualized Table. The code uses React Draggable for column resizing.

Upon some testing I realized that React Sortable HOC is interfering with React `Draggable`, probably because the Draggable component is inside `SortableElement`. I get undefined error messages to the `collection` property. The problem happens when a `Draggable` element is dragged over another `SortableElement`. This is because the `Draggable` component doesn't have a `collection` property, which is referenced by the `Manager.getActive` function.

The fix I came up is pretty simple, I just had to modify the `Manager.getActive` function to check if `this.active` is defined and only return `find` if `this.active` is not undefined. The `find` method would return `undefined` anyway if no item is found so there is no need for other code changes.

I'm preparing a pull request.

PR #171, opened 2017-03-26
Fix collection undefined when used with other drag and drop components

szabolcsx:
#170 
Returning find only when `this.active` is not undefined fixes the issue.

### DIFF
--- a/src/Manager.js
+++ b/src/Manager.js
@@ -21,11 +21,15 @@ export default class Manager {
   }
 
   getActive() {
-    return find(
-      this.refs[this.active.collection],
-      // eslint-disable-next-line eqeqeq
-      ({node}) => node.sortableInfo.index == this.active.index
-    );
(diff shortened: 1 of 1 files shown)


clauderic:
Hey @szabolcsx, while this does look like it fixes the issue you ran into, I the root of the issue must be somewhere else, as `getActive()` should never be called in the first place if `this.active` is undefined

szabolcsx:
I see. I will investigate further the problem.

clauderic:
Much appreciated, thank you 🙏

szabolcsx:
Hi,

I think I found the cause: the problem appeared when I started to move the cursor. I think the mouse move event was bubbling up, so I just had to do the following to stop that:

``` jsx
<Draggable
    { ...props }
    onDrag={event => event.stopPropagation()} />
```

PR #172, opened 2017-03-28
Fix ghost element hiding

amje:
Hi,

Currently, setting `visibility: hidden` on ghost element not guarantees its hiding. According to [spec](https://www.w3.org/TR/CSS22/visufx.html#visibility), descendants will be visible if they have `visibility: visible`. So additional `opacity: 0` should fix this.


amje:
@clauderic done

PR #173, opened 2017-03-29
hey! what's `react-infinite` doing in the `react-virtualized` example??

jedwards1211:
luckily it's just an accidental import


clauderic:
Hah! Good catch, thanks man!

PR #178, opened 2017-03-30
Remove 'pressThreshold' prop from WrappedComponent

RyanBertrand:
Hey - great use of HOC here!  Really love it 👍 

'pressThreshold' found it's way into the WrappedComponent's props since it was implemented in 0.6.0.  Small tweak to remove it to prevent polluting. 

Cheers,
Ryan


clauderic:
Hey @RyanBertrand, you're absolutely right, this was an oversight.

Thanks for submitting this ❤️

Issue #175, opened 2017-03-30
How is the node variable on onSortStart supposed to be used?


dumbmatter:
First of all, awesome library!

I'm trying to sort a table with variable width columns. It works, but the columns get smushed together for the row that is being dragged. Same thing happened using jQuery UI's sortable function in the past, which could be solved by manually setting the width of all the `td` nodes to their exact pre-sort values.

Looking at the API here, I saw the `onSortStart` function which gets a `node` variable, which I assumed to be the node of the dragged row, which I figured I could use for the same purpose. But I don't seem to be able to actually modify the dragged row. [Here is an example on JSFiddle](https://jsfiddle.net/46ew8jq1/), basically I have an `onSortStart` function like this:

        onSortStart({node}) {
            const children = node.childNodes;
            for (const child of children) {
                child.style.backgroundColor = 'green';
                child.style.width = child.offsetWidth;
            }
        }

So I'm attempting to manually set the width of all my columns, and also setting the background color to green just to make things visible. None of that applies to the dragged row, and the wrong row eventually turns green (after I stop dragging, the row that now appears where I started dragging is green).

So two questions I guess:

1. What are we supposed to be doing with this `node` variable, if it doesn't reference the dragged node?
2. Any ideas how I can achieve my goal of keeping the table columns properly sized?

dumbmatter:
Okay, I dug into the code a bit. Seems that the `node` variable is not what I want. If the `clonedNode` variable was also passed to `onSortStart`, I could do this and it'd work fine:

        onSortStart({clonedNode, node}) {
            const clonedChildren = clonedNode.childNodes;
            const children = node.childNodes;
            for (let i = 0; i < children.length; i++) {
                clonedChildren[i].style.backgroundColor = 'green';
                clonedChildren[i].style.width = `${children[i].offsetWidth}px`;
            }
        }

clauderic:
I'd be happy to accept a PR that would pass the clonedNode in `onSortStart`. As for the second alternative, I do think it would be a bit outside of the scope of this library, though it might be something worth documenting.

Going to close this issue, happy to keep discussing in a future PR 👍

PR #182, opened 2017-04-05
Expose clonedNode in onSortStart

dumbmatter:
This lets you access the `clonedNode` variable in the `onSortStart` callback, which is useful for manually applying styles to the item being moved, as discussed in #175.

I also added an example in the storybook. Delete the `onSortStart` in my story and see the difference, it goes from:

image

to

image

If you'd rather not have the story, I can delete it, leaving behind literally a one line PR :)


dumbmatter:
Anything I can do to move this along?

Issue #180, opened 2017-04-03
index missing in SortableElement


XeniaSiskaki:
Problem can be seen in https://jsfiddle.net/XeniaSiskaki/kajt284L/1/

I would like for each `SortableElement` to have an `x` icon so that the user can click on it and delete the element from the list. However, I cannot access the index in the `SortableElement`. Is that a bug or desired behavior?

clauderic:
Duplicate of https://github.com/clauderic/react-sortable-hoc/issues/128

This is an unfortunate side-effect of omitting `react-sortable-hoc` props from being passed down. I'll try to come up with a more suitable workaround, but in the meantime, you can do something like this:
https://jsfiddle.net/kajt284L/2/

XeniaSiskaki:
Thanks!

Issue #128, opened 2017-01-25
index is undefined in SortableElement v 0.4.10


tecnoloco:
Hello, I updated to version 0.4.10 and now when I try to access index prop inside my SortableElement it is undefined, in the meantime I will return to previous version where everything is fine.

```jsx
const SortableItem = SortableElement((props) => {
    const {name, fields, index} = props;
    
    if(fields.get(index).type == 'error'){
        return(
            <div className="file-upload-file-container file-container-error">
                {fields.get(index).file.name}
            </div>
        )

    } else{
        return (
            <div className="file-upload-file-container">
                {fields.get(index).file.name}
            </div>
        )
    }

});


const SortableList=SortableContainer(({fields, onDeleteItem})=>{
    return (
        <div className="">
            {fields.map((name, index, fields) => {
                return(
                    <SortableItem
                        key={`item-${index}`}
                        name={name}
                        index={index}
                        fields={fields}
                        >
                    </SortableItem>
                )
            })}
        </div>
    )
})
```

williamkhshea:
It looks like index is explicitly omitted in SortableElement. What is the best way to access index if I really need it?

clauderic:
Indeed, that's an unfortunate side-effect of https://github.com/clauderic/react-sortable-hoc/pull/100, which I've also run into myself a few times now and am unsure how to go about at this point.

For the moment, I've been dealing with it by passing down a different prop name like `i={index}`

Open to suggestions if you have ideas on the matter!

Robinfr:
Why not just stop omitting index?

stahlmanDesign:
I'm glad to find this thread because I couldn't figure out this bug. I am doing what @clauderic suggested, passing index as prop (which is `undefined` but component issues warning if you don't include it) and also under another name like indexTemp which I then assign to index in the child component.
```
index={ i }
indexTemp={ i }
```
in child component:
```
index={indexTemp}
```

AlastairTaft:
Seeing this issue for me, makes it unusable out the box for my case, going to try @clauderic's workaround.

derwaldgeist:
Came across this thread because I was wondering why the index isn't passed to the element. IMHO, this is quite important. In my use-case, each list item contains a "delete" button that enables the user to remove the item from the list. For this, I have to know in which item the delete button has been clicked. I'm already passing a `key` and an `index`, but both are "swallowed". Adding yet another index prop just seems to be overkill.

samelwitt:
[...]

PR #191, opened 2017-04-19
Pass disabled prop through to wrapped component

### LINKED ISSUE #190: Disabled prop no longer passed to WrappedComponent
Our implementation of a sortable list ca. 0.0.9 depended on the `disabled` prop being passed through to the wrapped child component, but it appears as of 0.6.2 this is no longer the case. Our workaround currently looks like this:

```js
const Item = SortableElement(({ child, itemClassName, disableSorting }) => {
  const classes = classnames('sortable-item', { 'sortable-item--disabled': disableSorting }, itemClassName)
  return <RecordListItem className={classes}>{child}</RecordListItem>
})

const List = SortableContainer(({ items, loaded, copy, itemClassName, disableSorting }) => {
  const renderChild = (child, index) => {
    const props = {
      key: `item-${index}`,
      itemClassName,
      index,
      child,
      disabled: disableSorting,
      disableSorting,
    }
    return <Item {...props} />
  }

  return (
    <RecordList loaded={loaded} copy={copy}>
      {items.map(renderChild)}
    </RecordList>
  )
})
```

Would be great to just use the `disabled` prop and not have a duplicate prop for this. Is there any reason why `disabled` is explicitly omitted from the passed-through props?

davidgoli:
Fixes #190

### DIFF
--- a/src/SortableElement/index.js
+++ b/src/SortableElement/index.js
@@ -86,7 +86,7 @@ export default function sortableElement(WrappedComponent, config = {withRef: fal
       return (
         <WrappedComponent
           ref={ref}
-          {...omit(this.props, 'collection', 'disabled', 'index')}
+          {...omit(this.props, 'collection', 'index')}
         />
       );
     }


clauderic:
This is an unfortunate side-effect of https://github.com/clauderic/react-sortable-hoc/issues/98

Unfortunately, I can't accept your PR, as I'd like to find a solution that would work for all of the cases mentioned in #128 and #180, while still ensuring that `react-sortable-hoc` consumes the props it uses.

See #128 and #180 for more context.

PR #196, opened 2017-04-23
Fixes "Cannot read property 'collection' of null" when distance set and a disabled element is moved

szabolcsx:
When the distance prop is set on a SortableContainer and a SortableElement is disabled, trying to move the element results in a "Cannot read property 'collection' of null" error being thrown.

This happens because when trying to sort a disabled element, the manager.active won't get set. In the SortableContainer's handleMove function, if distance is set, handlePress is executed even sorting was attempted on a disabled element.

### DIFF
--- a/src/Manager.js
+++ b/src/Manager.js
@@ -20,6 +20,10 @@ export default class Manager {
     }
   }
 
+  isActive() {
+    return this.active;
+  }
+
   getActive() {
     return find(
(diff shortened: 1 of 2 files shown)


Enet:
@clauderic, please, merge this PR. I have the same issue when trying to press on container instead of element.

clauderic:
Thanks for this PR, just published in `0.6.3`

Issue #58, opened 2016-09-23
useWindowAsScrollContainer not working in ie & Firefox


kapiljugnu:
ie version - 11.0.9600.18426
Firefox Version - 45.2.0

https://jsfiddle.net/vhcd9us3/1/

lapetr:
Firefox version 49.0 (Linux Ubuntu)

- http://clauderic.github.io/react-sortable-hoc/#/advanced/window-as-scroll-container
 - No auto-scroll on drag
 - bad sorting after drag e.g. "Item 1" and scrolling with mouse-wheel to next page
  - ![firefox49-bad-sort](https://cloud.githubusercontent.com/assets/9628720/19680032/635f5f3e-9aa4-11e6-9fb1-629b65f6f5a9.png)


(Chromium v. 53.0 Linux Ubuntu is OK )

asafigan:
It no longer works in chrome either

jimyaghi:
Strangely, only the delay 20ms example works in Chrome (ubuntu) when dragging into scroll overflow.

ThanosDi:
Any info on this? Also why and how is it only working on the 200ms example? 
http://clauderic.github.io/react-sortable-hoc/#/advanced/activate-onpress-200ms?_k=su6f6i

jimyaghi:
i've done pull request that sorts out all of these. can you take a look and see if it sorts out the problem for ya.

clauderic:
`useWindowAsScrollContainer ` was fixed in `0.7.0` thanks to #306, if you still experience issues with this, feel free to reopen

PR #198, opened 2017-04-28
Addressing issue #58 - useWindowAsScrollContainer bug

Jwhiles:
Currently useWindowAsScrollContainer doesn't work in firefox and internet explorer 

- I've added a function that checks whether the scroll top property is present on `document.documentElement` or `document.documentBody`. I think the code is slightly messy, but probably preferable to inlining an if statement into the existing ternary operator.

This works in Firefox - I'm not sure whether this also resolves the issue for internet explorer. I will try and determine and get access to a windows machine and test over the weekend.

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -105,12 +105,18 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
         useWindowAsScrollContainer,
       } = this.props;
 
+      const findScrollTop = () => {
+        if (this.document.body['scrollTop']) { return this.document.body; }
+        else if (this.document.documentElement['scrollTop']) { return this.document.documentElement; }
+        return;
+      };
+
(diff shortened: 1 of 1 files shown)


src/SortableContainer/index.js
```diff
@@ -105,12 +105,18 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
         useWindowAsScrollContainer,
       } = this.props;
 
+      const findScrollTop = () => {
+        if (this.document.body['scrollTop']) { return this.document.body; }
```
jimyaghi: how about: 
if('scrollTop' in this.document.body)

jimyaghi: Also, please note that in Chrome scrollTop is present in both but appears to be writeable ONLY in document.documentElement but not in document.body

clauderic:
Closed in favour of #306

PR #207, opened 2017-05-24
fix: keep origin transforms of helper

guan-wy:
If there are sortable helper with `transform` style, this `transform` style will be overwrite when dragging. This PR fixes this issue.

For example, as shown below

`SortableContainer` has `helperClass={ 'is-dragging' }`

`
.is-dragging {
    transform: rotate(2deg)
}
`

example


Brooooooklyn:
@clauderic friendly ping

Goues:
Hi, I tested this PR in a fork and there is a bug when original transform is none. Because then it append "none" to the final transform and dragging helper node doen't get translated.

PR #214, opened 2017-06-02
Fixed scrolling lists on touch screen

IlyaSkriblovsky:
Scrolling long sortable lists is broken on touchscreen because current master branch uses `e.clientX/Y` to get pointer coordinates and then uses them to calculate `delta`. And since `clientX/Y` are `undefined` for touch events, `delta` becames `NaN` and sorting is never cancelled even while scrolling.

This change is pretty trivial and uses already defined `getOffset` function to get pointer coordinates from an event.

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -143,10 +143,7 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
       }
 
       this._touched = true;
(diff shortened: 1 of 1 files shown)


fabioimpe:
@clauderic This works like a charm, please merge it!

clauderic:
This was fixed as part of #306

Issue #97, opened 2016-11-29
Table row style


GruFFix:
Hello, I applied react-sortable-hoc tables for sorting. When I drag a row, all styles have moved the element disappear. Please tell me how to be in this situation?
![image](https://cloud.githubusercontent.com/assets/16576776/20704267/65e57984-b640-11e6-9522-a4618748bc05.png)
p.s. Sorry for my english

clauderic:
That's probably because your css styling is tied to parent –> child relationship. The sorting helper gets cloned and appended to the `<body>` element, so if you're styling it with something like `table td`, it will be taken out of that context and all styles will disappear.

See this section of the README for more details: https://github.com/clauderic/react-sortable-hoc#item-disappearing-when-sorting--css-issues

The solution is to make sure your css is more specific, or re-apply that styling to the `helperClass`

GruFFix:
@clauderic Yes, I agree with you that the element <tr> element is designed with respect to <table>, but there <tr> can only exist within the <table> and its width is calculated with respect to the table

clauderic:
You won't be able to rely on those selectors. I know this is inconvenient, but there really is no way around it. In this specific example, you'd need to add an `.odd` and `.even` class to achieve that

johdah:
@clauderic I also have problems with table rows. When I drag it it loses the width since it's relies on the inherited width. Width feels like something the library should offer a possibility to set when it's moved. Now we only let some columns have a fixed width and let title and such be flexible.

PR #222, opened 2017-06-23
use scrollContainer as target for cloned node

frankleng:
@clauderic 
ran into #97, and thought `useWindowAsScrollContainer` would handle the use case where the node is cloned into the body.

so by default, clone to the container itself.


frankleng:
closing in favor of #194

Issue #231, opened 2017-07-01
Typescript support


ricokahler:
```
                                                      
                                                      
                                                      
             +------------------+                     
             | TYPESCRIPT USER  |                     
             |                  |                     
             |                  |                     
             |     +-------------------+              
             |     |            |      |              
             |     | ME ME ME   |      |              
             |     |            |      |              
             +-----|------------+      |              
                   |                   |              
                   |  REACT SORTABLE   |              
                   |  HOC USER         |              
                   |                   |              
                   +-------------------+              




```

This library is already has third party typings but I'd like to bring them in the repository for long term support.

There are two ways to do this:

1. bring the write type declarations in the repository
2. port the whole project to typescript

I'm actually in favor of option 2.

OP, If you're not a typescript user, maybe I can convince you?

Let me know your thoughts! Thanks for the great library 👍 🥇

clauderic:
Hey @ricokahler,

I'd love to look into porting this library to TypeScript, I already use it a fair amount for work related projects.

Having said that, I don't have the bandwidth to convert the library myself. If this is something you feel strongly about, a PR to that effect would be most welcome 👍

dlindenkreuz:
Wow guys, option 2 is quite an endeavour.

TS definition files would be perfectly alright as well, I'm gonna check for myself if this is worth taking a shot.

dlindenkreuz:
Done ✅

This should solve the issue for now. The declaration file is all I needed to work with react-sortable-hoc in a large-scale Typescript project.

PR #234, opened 2017-07-04
[WIP] Typescript port

ricokahler:
# This is very much a work in progress.

In fact, this code currently doesn't even compile. The reason I'm opening the pull request now is to ask questions and get feedback early before I change almost everything 😅 **though the exposed API won't change so nothing will break** if this eventually accepted.

@clauderic could you answer the questions below by the number when you have the time?

## Questions:

1. **How are touch events handled?** It seems like we're adding events to the raw container but there is no use of [`TouchEvent.changedTouches`](https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent/changedTouches). For the event handler `handleStart`, I've typed the event as [`e: MouseEvent | TouchEvent`][0] but the touch events *don't* have `e.pageX` and `e.pageY` defined. [This is how I'm getting `pageX` and `pageY` from the touch events][1]. **Is this change okay?** I suspect we might have issues if both a click event and touch event are fired but this can be confirmed later.

[0]: https://github.com/ricokahler/react-sortable-hoc/blob/f9e281295b5e5f5d984746d83e3e9f82966c1db1/src/SortableContainer/index.tsx#L207
[1]: https://github.com/ricokahler/react-sortable-hoc/blob/f9e281295b5e5f5d984746d83e3e9f82966c1db1/src/SortableContainer/index.tsx#L217-L225

2. **What is the type for [`Manager.prototype.getActive`][2]?** I see that you're taking `HTMLElement`s and adding the property `sortableInfo` to them so [I've created and interface for that][4]. I also see the property `refs` on the class `Manager` holds these elements. I've typed this property as [`refs: { [collection: string]: SortableNode[] }`][5]. Now my question lies with how you use [`manager.getActive()` from `handlePress`][6]. My interface for `Manager.refs` doesn't agree with your implementation because [these `refs` don't have a `collection` or `node` property][7]. **What is the correct type for `refs`? Is it correct to use [`this.manager.getActive()` here?][8]**

[2]: https://github.com/ricokahler/react-sortable-hoc/blob/f9e281295b5e5f5d984746d83e3e9f82966c1db1/src/Manager.ts#L39
[3]: https://github.com/ricokahler/react-sortable-hoc/blob/f9e281295b5e5f5d984746d83e3e9f82966c1db1/src/SortableContainer/index.tsx#L331
[4]: https://github.com/ricokahler/react-sortable-hoc/blob/f9e281295b5e5f5d984746d83e3e9f82966c1db1/src/Manager.ts#L3-L9
[5]: https://github.com/ricokahler/react-sortable-hoc/blob/f9e281295b5e5f5d984746d83e3e9f82966c1db1/src/Manager.ts#L12
[6]: https://github.com/ricokahler/react-sortable-hoc/blob/f9e281295b5e5f5d984746d83e3e9f82966c1db1/src/SortableContainer/index.tsx#L319-L331
[7]: https://github.com/ricokahler/react-sortable-hoc/blob/f9e281295b5e5f5d984746d83e3e9f82966c1db1/src/SortableContainer/index.tsx#L331
[8]: https://github.com/ricokahler/react-sortable-hoc/blob/f9e281295b5e5f5d984746d83e3e9f82966c1db1/src/SortableContainer/index.tsx#L320

---

Let me know your thoughts, and if you live in the states, happy fourth!


ricokahler:
https://github.com/clauderic/react-sortable-hoc/issues/231

src/Manager.ts
```diff
@@ -0,0 +1,58 @@
+import { find, sortBy } from 'lodash';
+
+export interface SortableNode extends HTMLElement {
+  sortableInfo: {
+    index: number,
+    collection: string,
+    manager: Manager,
+  }
+}
```
clauderic: Nice 👍

examples/basic.js
```diff
@@ -1,10 +1,10 @@
-import React, {Component} from 'react';
-import {render} from 'react-dom';
-import {SortableContainer, SortableElement, arrayMove} from 'react-sortable-hoc';
+import React, { Component } from 'react';
+import { render } from 'react-dom';
```
clauderic: Small nitpick: Can we revert all these to have no spaces? Just a personal preference wrt linting 😅

ricokahler: @clauderic yeah we can do that. I have no preference. VS Code is my main editor and I use the shortcut to format code and that was the result. I just have to change my vs code settings somewhere.

clauderic: You can use [prettier](https://github.com/prettier/prettier) for all linting, it's what I used with the repo, with these configs:

```
 "prettier.singleQuote": true,
 "prettier.trailingComma": "es5",
 "prettier.bracketSpacing": false,
```

clauderic:
Moving this over to #250

PR #239, opened 2017-07-09
improve handleStart lag by removing unnecessary update/render cycle

jakedowns:
small tweak: moved .sorting to a local property instead of a this.setState call
could make an immutable scoped var or something, but this works for now.

* * *

I was experiencing some lag on .handleStart due to the setState call on the container
(~30ms with 57 items / ~165ms with 351 items)

by moving this field to a class-level property, rather than Component.state, handleStart now executes in ~1-5ms for me (even with 350 items)

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -32,6 +32,7 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
       );
 
       this.state = {};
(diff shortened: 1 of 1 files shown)


matthewk:
I have an issue with the current sortable HOC that I believe would be fixed by this, is there anything stopping this from being merged?

Issue #242, opened 2017-07-13
Overflow bug while dragging an item upwards in a grid


InterAl:
The bug can be seen in your examples website:

http://clauderic.github.io/react-sortable-hoc/#/basic-configuration/grid?_k=pa2nzs

Try dragging an element from the second row to the first row.

The first row, while dragging into it, contains 4 elements instead of 3.

After dropping, the container re-adjusts and the first line contains 3 elements as before.

![image](https://user-images.githubusercontent.com/8208975/28157159-43427a64-67be-11e7-8dcf-702e4bef7bf2.png)

clauderic:
Fixed in #479

PR #243, opened 2017-07-13
Fixed a right-overflow bug in a grid setup

InterAl:
https://github.com/clauderic/react-sortable-hoc/issues/242


cusxio:
@clauderic any chance we can get this merge to master? I can confirm that this bug is indeed valid, and this PR fixes it.

@InterAl thank you !

PR #244, opened 2017-07-14
Update link to `shouldCancelStart` function definition

anaumov:
Update link to `shouldCancelStart` function definition, line number was wrong.


clauderic:
Thanks 😄

PR #248, opened 2017-07-18
IE 11 fixes for react-sortable-hoc

humiston:
Added window.pageYOffset and window.pageXOffset to updatePosition and animateNode to allow proper drag and drop animation for Internet Explorer 11

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -520,8 +520,8 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
         y: offset.y - this.initialOffset.y,
       };
       // Adjust for window scroll
-      translate.y -= ((window.scrollY || window.pageYOffset) - this.initialWindowScroll.top);
-      translate.x -= ((window.scrollX ||  window.pageXOffset)- this.initialWindowScroll.left);
+      translate.y -= (window.pageYOffset - this.initialWindowScroll.top);
+      translate.x -= (window.pageXOffset - this.initialWindowScroll.left);
 
       this.translate = translate;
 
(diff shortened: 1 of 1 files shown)


src/SortableContainer/index.js
```diff
@@ -513,15 +513,15 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
...
       const offset = this.getOffset(e);
       const translate = {
         x: offset.x - this.initialOffset.x,
         y: offset.y - this.initialOffset.y,
       };
       // Adjust for window scroll
-      translate.y -= (window.scrollY - this.initialWindowScroll.top);
-      translate.x -= (window.scrollX - this.initialWindowScroll.left);
+      translate.y -= (window.scrollY - this.initialWindowScroll.top) || window.pageYOffset;
+      translate.x -= (window.scrollX - this.initialWindowScroll.left) || window.pageXOffset;
```
clauderic: Looks to me like those two statements are not equivalent. 

The first one will take into account the diff between the current `scrollY` and `scrollX` values vs the initial scroll value

humiston: I've updated this pr with the correct assignment of our initial scroll values so our diff is accurate when using IE 11.

clauderic:
Just published `0.6.6` with this fix, thanks again for your contribution 😄

mccambridge:
@clauderic @humiston I'm running storybook with this fix, and I still see the issue with IE11. Both in my code and in your Drag Handle storybook example. Items aren't draggable at all.

timothyallan:
It's because IE11 doesn't support Array.From() which is being used in the code. I just ran into this myself.

PR #255, opened 2017-07-27
Set draggable helper to maximum z-index

eshork:
Fixes visual anomaly when SortableContainer is used within material-ui modal dialog (probably other similar situations as well). The draggable element has a lower z-index than the modal dialog, so it becomes hidden behind the dialog.

This solves the issue by setting z-index to the maximum positive value, which should ensure it is always drawn on top of all other elements.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-sortable-hoc",
-  "version": "0.6.6",
+  "version": "0.6.7",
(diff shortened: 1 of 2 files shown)


clauderic:
This should be handled in your application, most applications will not necessarily want to default to the highest maximum z-index and will instead want to decide for themselves what the z-index of the helper should be.

This can be achieved using the `helperClass` prop

PR #260, opened 2017-08-07
Set contentWindow default ourselves to avoid Jest serialization

### LINKED ISSUE #249: Out of memory when snapshotting react-sortable-hoc
Jest/Node crashes with a memory leak when trying to snapshot a component that contains an element connected with react-sortable-hoc.

The Jest developers stated that this is not an issue with the snapshotting, but with the library.

It makes it very hard to write proper tests for anything that involves a sortable.

Simple test case:

```js
import React from 'react';
import { shallow } from 'enzyme';
import serializer from 'enzyme-to-json/serializer';
import { SortableContainer } from 'react-sortable-hoc';

expect.addSnapshotSerializer(serializer);

const Component = () => (
  <div />
);

const Sortable = SortableContainer(Component);

test('jest crashes on snapshot creation', () => {
  const component = shallow(
    <div>
      <Sortable />
    </div>
  );

  expect(component).toMatchSnapshot();
});
```
Jest 20.0.4, Node 8.1.4 on macOS 10.12.5

The actual error (which Google tells me is basically out of memory): `RangeError: Invalid string length`

cameronmcefee:
This fixes #249

Jest serializes all props in snapshots. The current code sets window by default, which is a huge object. When Jest tries to serialize it, it typically throws `RangeError: Invalid string length` because the process runs out of memory. If you're lucky enough to actually get a test to pass, the snapshot will be around 250mb.

This switches from using defaultProps to setting the default ourselves, so in the test environment there won't be a prop value to serialize.

I confirmed the commonjs version fixes the issue. I'm not sure how to test the es6/umd versions. `npm test` passes.

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -44,7 +44,6 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
       distance: 0,
       useWindowAsScrollContainer: false,
       hideSortableGhost: true,
-      contentWindow: typeof window !== 'undefined' ? window : null,
       shouldCancelStart: function(e) {
         // Cancel sorting if the event target is an `input`, `textarea`, `select` or `option`
         const disabledElements = ['input', 'textarea', 'select', 'option', 'button'];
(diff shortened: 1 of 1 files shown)


clauderic:
Just published this fix with `0.6.7`

cameronmcefee:
❤️ Thanks, this component has saved me much time. I'm happy to have contributed.

PR #272, opened 2017-09-06
add cloneComputedStyles option

PsiXdev:
Add option to clone all computed styles from moved element to it copy.


heyarne:
Any chance of getting this merged? This would solve a lot of problems with elements that depend on specific parents for their appearance, which is common for many CSS frameworks.

Matrix89:
Will this ever get merged?

PR #276, opened 2017-09-08
Update Code Sponsor link position

cavneb:
@clauderic When the link is found at the bottom of the README, it is more effective in helping funding this project.


clauderic:
Hmm, interesting. Is this something you've A/B tested with different README lengths?

Let's see how that goes.

PR #301, opened 2017-10-27
Make bundle size smaller

374632897:
* use ramda instead of lodash
* move prop-types to peerDependencies


blord-fullscreen:
How does using ramda utilities make the bundle size smaller? Can you show a comparison table, compressed and uncompressed, etc?

374632897:
@blord-fullscreen Because lodash/find and lodash/sortby will use other methods of lodash, looks like you just use these two methods , but others will be referenced.  
```js
// lodash/find
var createFind = require('./_createFind'),
    findIndex = require('./findIndex');  

// lodash/_createFind
var baseIteratee = require('./_baseIteratee'),
    isArrayLike = require('./isArrayLike'),
    keys = require('./keys');
// ...
```
After we used ramda instead,  the uncompressed bundle size reduced from 151kb to 46kb.   

|  | bundle size(uncompressed)|finally size of find & sortBy|percent|
|---|-------|---------|------------|
|lodash|151.38kb|104.7kb|69.16%|
|ramda|46.42kb|7.96kb|17.14%|

clauderic:
Closing in favour of #374

PR #309, opened 2017-11-07
add custom ghost parent element prop instead of always using document.body

daviestar:
fixes https://github.com/clauderic/react-sortable-hoc/pull/194#issuec…omment-297588777


clauderic:
This has been possible for quite some time now using the `helperContainer` prop

PR #326, opened 2017-12-03
Stub out Virtualized Grid Example

blord-fullscreen:
Hi @clauderic, I am trying to get this working with react-virtualized `<Grid>`. I  am running into an issue I'm hoping you can spot and address. In hopes of tempting you to help me out, I've stubbed out an illustrative example in the storybook demo page 😸  which I hope can be tweaked and added to the repo for future purposes.

Basically, everything works, except that drag-scrolling down long distances generates errors in the console. The error is easy to reproduce. Run the demo locally, click on the . "Virtualized Grid" example, and drag scroll down, hard and fast. The error happens here:

https://github.com/clauderic/react-sortable-hoc/blob/master/src/SortableContainer/index.js#L681

I would be extremely grateful for a fix to this. Thanks for an awesome library.

### DIFF
--- a/src/.stories/index.js
+++ b/src/.stories/index.js
@@ -8,6 +8,7 @@ import VirtualList from 'react-tiny-virtual-list';
 import {
   defaultTableRowRenderer,
   Column,
+  Grid,
   Table,
   List,
 } from 'react-virtualized';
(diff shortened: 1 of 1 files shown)


jfrolich:
It looks to work for me. The only issue here is that the ordering seems to be unstable, if you drag and pause, the items shift. This does not happen if you keep the mouse 'wiggling'. Example below (gif):

chaomao:
https://codesandbox.io/s/dazzling-rhodes-q6jku, a simple demo

PR #352, opened 2018-03-12
Text wraps on dragging

AdriatikDushica:
In this example i'm dragging the first item and the text inside automatically wraps:

example

The problem is that the node.offsetWidth returns 131 pixels instead of 131.16.

It can be solved by replacing node.offsetWidth and node.offsetHeight with node.getBoundingClientRect().width and node.getBoundingClientRect().height

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -55,8 +55,8 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
       lockToContainerEdges: false,
       lockOffset: '50%',
       getHelperDimensions: ({node}) => ({
-        width: node.offsetWidth,
-        height: node.offsetHeight,
+        width: node.getBoundingClientRect().width,
+        height: node.getBoundingClientRect().height,
       }),
     };
 
@@ -584,8 +584,8 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
       for (let i = 0, len = nodes.length; i < len; i++) {
         const {node} = nodes[i];
         const index = node.sortableInfo.index;
-        const width = node.offsetWidth;
-        const height = node.offsetHeight;
+        const width = node.getBoundingClientRect().width;
+        const height = node.getBoundingClientRect().height;
         const offset = {
           width: this.width > width ? width / 2 : this.width / 2,
           height: this.height > height ? height / 2 : this.height / 2,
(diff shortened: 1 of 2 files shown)


src/SortableContainer/index.js
```diff
@@ -584,8 +584,8 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
       for (let i = 0, len = nodes.length; i < len; i++) {
         const {node} = nodes[i];
         const index = node.sortableInfo.index;
-        const width = node.offsetWidth;
-        const height = node.offsetHeight;
+        const width = node.getBoundingClientRect().width;
+        const height = node.getBoundingClientRect().height;
```
hledley: I don't think `getBoundingClientRect` is "free". Thoughts on writing this
```
const {width, height} = node.getBoundingClientRect()
```
?

Could be written similarly above.

hledley:
👍 noticed this problem as well, same solution.

hledley:
One note if this is blocking you -- you can pass a custom `getHelperDimensions` with your updated code and you should be mostly fine

src/SortableContainer/index.js
```diff
@@ -55,8 +55,8 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
       lockToContainerEdges: false,
       lockOffset: '50%',
       getHelperDimensions: ({node}) => ({
-        width: node.offsetWidth,
-        height: node.offsetHeight,
+        width: node.getBoundingClientRect().width,
+        height: node.getBoundingClientRect().height,
       }),
```
gunn: Similar to @hledley's suggesion, this should only call `getBoundingClientRect` one
```suggestion
      getHelperDimensions: ({node}) => node.getBoundingClientRect()),
```

src/SortableContainer/index.js
```diff
@@ -55,8 +55,8 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
       lockToContainerEdges: false,
       lockOffset: '50%',
       getHelperDimensions: ({node}) => ({
-        width: node.offsetWidth,
-        height: node.offsetHeight,
+        width: node.getBoundingClientRect().width,
+        height: node.getBoundingClientRect().height,
```
gunn: Similar to @hledley's suggestion, this should only call `getBoundingClientRect` once
```suggestion
      getHelperDimensions: ({node}) => node.getBoundingClientRect()),
```

Issue #254, opened 2017-07-27
helperClass isn't applied to currently active element in IE


mccambridge:
The `helperClass` I specified in `SortableList` works great in other browsers. But it's not being applied to the element in IE11.

mccambridge:
Looks like your distributable uses array spread without transpiling to populate the classList string.

```this.helper.classList.add(...helperClass.split(' '));```

So you're taking the helper class string, converting it to an array, and dropping it back into a string via spread?

dcorb:
Same issue here, IE11 doesn't add helperClass to the dragging element

dcorb:
IE11 only loads the first class of helperClass (for example helperClass="foo bar", will load only "foo")

I used create-react-app and copy/pasted this polyfill https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from#Polyfill  to avoid another IE11 error about Array.from not supported

decafdennis:
This is due to `Element.classList.add` not supporting multiple arguments in IE11, as noted in the browser compatibility table here: https://developer.mozilla.org/en-US/docs/Web/API/Element/classList

I was able to make v0.8.3 work using [`classlist-polyfill`](https://www.npmjs.com/package/classlist-polyfill).

clauderic:
This is now fixed and no longer requires a polyfill https://github.com/clauderic/react-sortable-hoc/blob/master/src/SortableContainer/index.js#L341

PR #358, opened 2018-03-23
[#254] Fix a bug where the helperClass isn't correctly applied

HighCoordination:


decafdennis:
This is due to `Element.classList.add` not supporting multiple arguments in IE11, as noted in the browser compatibility table here: https://developer.mozilla.org/en-US/docs/Web/API/Element/classList

You should probably use [`classlist-polyfill`](https://www.npmjs.com/package/classlist-polyfill).

clauderic:
This was fixed in a separate PR, see https://github.com/clauderic/react-sortable-hoc/blob/master/src/SortableContainer/index.js#L341

PR #384, opened 2018-05-29
Fix TouchEvent undefined message

AlekzandrG:
Fix TouchEvent undefined check for browsers that not support this functionality


clauderic:
This was already fixed in `0.8.2`. Not sure what exactly the benefit of this code over what’s already there would be

AlekzandrG:
Is not fixed at all in Firefox and Safari the console throws an error related to event instanceof window.TouchEvent as Is exposed un issue 383

PR #396, opened 2018-07-11
Replace imports from react-dom

damilaufer:
react-dom exports using CommonJs modules. When building with Rollup, I get the error below because the way the imports are written are not compatible with ES6 modules.

The error when building with Rollup and Babel 6:
[!] Error: 'findDOMNode' is not exported by node_modules/react-dom/index.js

More info: https://blog.kentcdodds.com/misunderstanding-es6-modules-upgrading-babel-tears-and-a-solution-ad2d5ab93ce0

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Usage
 
 ```js
 import React, {Component} from 'react';
-import {render} from 'react-dom';
+import ReactDOM from 'react-dom';
 import {SortableContainer, SortableElement, arrayMove} from 'react-sortable-hoc';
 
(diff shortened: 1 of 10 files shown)


ssinghi:
Shouldn't this be changed in react-dom to not use CommonJS modules, but rather export using ES6 modules?

damilaufer:
That'll be the best solution. But it'll take some time (they will like to move all their libraries to ES6 modules).
Until then, this PR will take care of the issue for all the users of this package. It's a small change, that doesn't affect size, performance or tests.

PR #417, opened 2018-08-16
Add a function to get the DOM node to mount the helpers at.

jhicken:
I suppose I should add documentation. ...

First though does this seem reasonable?


clauderic:
There is the `helperContainer` prop that you can now use for this purpose

Issue #395, opened 2018-07-10
Canvas element not successfully cloned


Benjaki2:
When the clone is appended to the `<body>` it doesn't draw the old canvas image onto a new canvas.

nurpax:
This seems to be because the [Node.cloneNode](https://developer.mozilla.org/en-US/docs/Web/API/Node/cloneNode) doesn't retain the canvas contents.  Kind of a bummer.

It'd be possible to work-around this by cloning it so that we copy the rendered shapes into an image and then blit that image into the new canvas element.  Found some jquery code to do that here: https://gist.github.com/adouglas/5266778

I'd need this in my project too.

nurpax:
Here's how the canvas misbehaves in the react-sortable-hoc main example (the one at the project root):

![canvas-hoc-buggy](https://user-images.githubusercontent.com/297823/44622401-56e78200-a8c0-11e8-965b-470b7d2ed4d5.gif)

I found a way to fix this.  I'll post a PR here soon.

Here's how it works after my fix:

![canvas-hoc-fixed](https://user-images.githubusercontent.com/297823/44622408-754d7d80-a8c0-11e8-8b8f-8b43ecefd4c0.gif)

nurpax:
https://github.com/nurpax/react-sortable-hoc/tree/canvas-drawimage

nurpax:
My PR is ready.  Just seems like PRs are not going anywhere.

See https://github.com/clauderic/react-sortable-hoc/pull/421#issuecomment-416328346

I'm using my own fork of react-sortable-hoc in my project as I really needed this fix.

PR #341, opened 2018-02-11
when creating the ghost element for drag effect, copy the canvas content as well

ssinghi:
Fixes issues #330.

PR #421, opened 2018-08-25
Clone canvas contents with drawImage (#395)

nurpax:
When DOM nodes are cloned with Node.cloneNode, the contents of
<canvas> elements are lost.  This makes canvas elements lose their
contents while dragged.

On losing <canvas> contents: "Cloning a node copies all of its
attributes and their values, including intrinsic (in–line)
listeners. It does not copy event listeners added using
addEventListener() or those assigned to element properties
(e.g. node.onclick = fn). _Moreover, for a <canvas> element, the
painted image is not copied._"
(https://developer.mozilla.org/en-US/docs/Web/API/Node/cloneNode,
emphasis mine)

To fix this, copy the painted image from the old canvas into the new
cloned canvas with drawImage.

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -289,6 +289,7 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
           left: window.pageXOffset,
         };
 
+        const canvases = node.querySelectorAll('canvas');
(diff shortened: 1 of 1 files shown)


tzachyrm:
@nurpax this looks like a duplicate of #341.
Sadly it doesn't seem like PRs get merged in this repo...

Issue #428, opened 2018-10-08
Different height items cause wrong sorting animation order


adrianhelvik:
Hi. I have noticed an issue. When items have different heights, the wrong item moves. In the repro, if you attempt to move *0* slowly down, you will notice that *2* moves before *1*

Repro: https://codepen.io/anon/pen/EdNGOz

adrianhelvik:
When dropping the sortable element, the sorting works as expected, so this is is related to the move events.

adrianhelvik:
Have a look at:

SortableContainer/index.js at line 545
```
        const offset = {
          width: this.width > width ? width / 2 : this.width / 2,
          height: this.height > height ? height / 2 : this.height / 2,
        };
```

and SortableContainer/index.js at line 659

```
        } else if (this.axis.y) {
          if (
            index > this.index &&
            (sortingOffset.top + windowScrollDelta.top) + offset.height >= edgeOffset.top
          ) {
            translate.y = -(this.height + this.marginOffset.y);
            this.newIndex = index;
          } else if (
            index < this.index &&
            (sortingOffset.top + windowScrollDelta.top) <= edgeOffset.top + offset.height
          ) {
            translate.y = this.height + this.marginOffset.y;
            if (this.newIndex == null) {
              this.newIndex = index;
            }
          }
        }
```

and in that piece of code, this one:

```
            (sortingOffset.top + windowScrollDelta.top) + offset.height >= edgeOffset.top
```

If I remove `offset.height` the animations occur in the correct order, but only start when the top edge of the dragged element touches the element below.

# Cause and solution

The offset should be calculated based on the dragged node and not the target. When the height of the dragged element (this.height) is larger than the height of the element being hovered over (height), the ternary condition kicks in and the bug is producted.

```
        const offset = {
          width: this.width > width ? width / 2 : this.width / 2,
          height: this.height > height ? height / 2 : this.height / 2,
        };
```

The solution for this error condition is to change offset to be simply:

```
        const offset = {
          width: this.width / 2,
          height: this.height / 2,
        };
```

Is there a reason why the dimensions of the target node is used here? If that is the case, would it be possible to override this with some configuration? Otherwise it can be fixed.

adrianhelvik:
Another update: This produced other issues. I closed the pull request and will look further into the issue. Help is appreciated.

evanc:
[...]

PR #430, opened 2018-10-09
Solves #428 - Different height items cause wrong sorting animation order

adrianhelvik:
Please consider why the target node was used as a reference node for the height. It does not make sense to me, but there might be something I am overlooking.


adrianhelvik:
I will create a new branch with only the relevant change.

PR #415, opened 2018-08-14
Fix bug: sortable items disappear in full screen mode

TungAng:
See issue #414 for more details about the bug.

Issue #414, opened 2018-08-14
Sortable items disappear in full screen mode


TungAng:
When you try to drag the SortableElement in full-screen mode, it will disappear from the screen and only reappear after finish sorting.

![react-app](https://user-images.githubusercontent.com/24927510/44070251-f668804e-9fac-11e8-95e5-b19198bab6fe.gif)

Code:
```
import React, { Component } from "react";
import { render } from "react-dom";
import {
  SortableContainer,
  SortableElement,
  arrayMove
} from "react-sortable-hoc";

const SortableItem = SortableElement(({ value }) => <li>{value}</li>);

const SortableList = SortableContainer(({ items }) => {
  return (
    <ul>
      {items.map((value, index) => (
        <SortableItem key={`item-${index}`} index={index} value={value} />
      ))}
    </ul>
  );
});

const openFullscreen = elem => {
  if (elem.requestFullscreen) {
    elem.requestFullscreen();
  } else if (elem.mozRequestFullScreen) {
    /* Firefox */
    elem.mozRequestFullScreen();
  } else if (elem.webkitRequestFullscreen) {
    /* Chrome, Safari and Opera */
    elem.webkitRequestFullscreen();
  } else if (elem.msRequestFullscreen) {
    /* IE/Edge */
    elem.msRequestFullscreen();
  }
};

class SortableComponent extends Component {
  state = {
    items: ["Item 1", "Item 2", "Item 3", "Item 4", "Item 5", "Item 6"]
  };

  onSortEnd = ({ oldIndex, newIndex }) => {
    this.setState({
      items: arrayMove(this.state.items, oldIndex, newIndex)
    });
  };

  handleFullscreenButtonClick = () => {
    openFullscreen(this.sortableList);
  };

  render() {
    return (
      <div>
        <div
          ref={elem => {
            this.sortableList = elem;
          }}
        >
          <SortableList items={this.state.items} onSortEnd={this.onSortEnd} />
        </div>
        <button onClick={this.handleFullscreenButtonClick}>
          click to fullscreen
        </button>
      </div>
    );
  }
}

render(<SortableComponent />, document.getElementById("root"));

```

overfed:
the same thing if we want to use sortable items in popover.

clauderic:
This is due to the fact that by default, the clone of the SortableElement is appended to the document.body. You can change this behaviour by using the `helperContainer` prop

PR #434, opened 2018-10-15
Add support for FullScreen API

garth:
See #414 and #415


garth:
Until support for the FullScreen API is merged/added in to the official lib, a modified version of `react-sortable-hoc` with only this PR added has been published to npm.

```
yarn add react-sortable-hoc-fullscreen
```

clauderic:
I believe you could use the newly added `helperContainer` prop instead to achieve the same functionality. Going to close this for now unless there is a strong case to be made to support this natively

garth:
@clauderic no, doesn't work. To fix, helperContainer would have a to accept a function that is evaluated when the drag begins.

garth:
In case someone finds this issue, since version 1.5.2 you can do the following to get fullscreen support:

```tsx
<SortableList
  helperContainer={() =>
    document['fullscreenElement'] ||
    document['webkitFullscreenElement'] ||
    document['mozFullScreenElement'] ||
    document['msFullscreenElement'] ||
    document.body
  }
/>
```

clauderic:
Thanks for sharing back your solution @garth

PR #442, opened 2018-11-19
Fix the dimension calculation

alexandcote:
Fix the dimension calculation when helperClass change the size of the node.

### After

I tried to add the helperClass only on the clone element but we can't because the hidden element will still take the same height.

Let me know if you need more detail ✌️


alexandcote:
After discussion, we will try another approach so I will close this PR.

PR #447, opened 2018-11-22
Fix a bug when you use SortableHandle and distance prop.

alexandcote:
I found an issue when you set a distance and you use `SortableHandle`.
After you click on the handle, you are able to drag the element from everywhere inside the `li`.

| Before        | After           |
| ------------- |:-------------:|
| |  |

So the main problem comes from [here](https://github.com/clauderic/react-sortable-hoc/blob/master/src/SortableContainer/index.js#L179). During the handleStart we set the manager.
But inside the `handleEnd` we never unset the manager if the distance prop was set. You can check the code [here](https://github.com/clauderic/react-sortable-hoc/blob/master/src/SortableContainer/index.js#L227). It's true that we shouldn't call `clearTimeout(this.pressTimer);` when you had set a distance because you can't set both properties (distance | pressDelay). 

Feel free to ask me for changes if you find a better approach ✌️

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -225,18 +225,16 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
     };
 
     handleEnd = () => {
-      const {distance} = this.props;
-
       this._touched = false;
-
-      if (!distance) {
-        this.cancel();
-      }
+      this.cancel();
     };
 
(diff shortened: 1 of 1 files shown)


clauderic:
Just shipped `react-sortable-hoc@0.8.4` with this bug fix, thanks for submitting this PR

PR #457, opened 2018-12-13
fix: Sortable grid first row problem

praya:
this merge resolves this issue:
https://github.com/clauderic/react-sortable-hoc/issues/242


praya:
Hello guys!
Why don't you answer me?

PR #458, opened 2018-12-14
Implement a new callback before we create the ghost beforeSortStart.

alexandcote:
This PR adds a callback that is invoked before we create the ghost. This can be really useful if, for example, you would like to collapse an item during sorting.

We can't use the `helperClass` in this situation because the height of the element changed. We also can't use the `onSortStart` callback because the ghost is already created.

Let me know if you have quesitons ✌️


src/SortableContainer/index.js
```diff
@@ -207,7 +208,7 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
     handleMove = event => {
       const {distance, pressThreshold} = this.props;
 
-      if (!this.state.sorting && this._touched) {
+      if (!this.state.sorting && this._touched && !this._beforeSortStartIsCalled) {
```
alexandcote: We need this check because if the user drag before the promise is resolved, the drag is cancel

src/SortableContainer/index.js
```diff
@@ -250,11 +251,20 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
...
+          beforeSortStart,
           onSortStart,
           useWindowAsScrollContainer,
         } = this.props;
+
         const {node, collection} = active;
         const {index} = node.sortableInfo;
+        
+        if (beforeSortStart) {
+          this._beforeSortStartIsCalled = true;
```
alexandcote: Open to a better name/approach for this.

src/SortableContainer/index.js
```diff
@@ -241,7 +242,7 @@ export default function sortableContainer(WrappedComponent, config = {withRef: f
       }
     };
 
-    handlePress = event => {
+    handlePress = async event => {
```
beefchimi: I bring this up only because I don't see `async/await` else where in the code base – is this within support range for the library?

clauderic:
Published as a beta, install `react-sortable-hoc@0.9.0-beta` to test it out

utkarshsaxena93:
@clauderic  working as expected 👍

Issue #444, opened 2018-11-20
Remove unsafe lifecycle methods


rwrobe:
SortableElement uses componentWillReceiveProps which is deprecated and will be removed in 17

lavebug:
@rwrobe have you figured what to use instead of componentWillReceiveProps(nextProps)? Any examples?

clauderic:
Fixed in #472

PR #472, opened 2019-01-08
fix: updated the behaviour of disabled elements

### LINKED ISSUE #448: Unable to drag items before or after disabled items
The issue is repro here:
https://jsfiddle.net/acupy/d67r8fjr/58/

Moving 'Item 1' after 'Item 3' or moving 'Item 4' before 'Item 2' does not seem to work.

Could you help me with this please?

clauderic:
Fixes #448 #444 

kapture 2019-01-08 at 13 15 39

### DIFF
--- a/src/.stories/Storybook.scss
+++ b/src/.stories/Storybook.scss
@@ -46,6 +46,11 @@
   font-weight: 400;
 }
 
+.disabled {
+  cursor: not-allowed;
+  opacity: 0.5;
+}
+
 // Drag handle
 .handle {
   display: block;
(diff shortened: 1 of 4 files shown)


Zodiase:
We had to upgrade react-sortable-hoc to fix another bug but this change seems to have just broke our app. Is there an option to achieve the now "fixed" behavior, so a "disabled" item doesn't participate in sorting at all?

clauderic:
@Zodiase you can achieve the same behaviour as before by conditionally rendering a version of your component that is not enhanced by `SortableElement`, something like this (pseudo-code):

```js
class Item extends React.Component {
  ...
}

const SortableItem = sortableElement(Item);

class MyComponent extends React.Component {
  render() {
    const {items} = this.props;

    return (
      <SortableComponent>
        {items.map(item => item.disabled ? <Item /> : <SortableItem />)}
      </SortableComponent>
    );
  }
}

PR #482, opened 2019-01-11
feat: add disableAnimation prop

krzaczek:
Hi,

In one of our projects we had to disable elements "animation" (switching places) until onSortEnd. We added new prop `disableAnimation` that does just that.

We use `onSortOver` to mark the position where the element will be dropped.

This allows us to achieve something like this:
disableAnimation

Same with animations enables:
nodisableAnimation


clauderic:
Hey @krzaczek, thanks for taking the time to submit this PR. 

Unfortunately I don't really see this as a common use-case, and for that reason I don't think it fits within the base use-cases that this library aims to support. Have you considered using something like `react-dnd` for this use-case instead?

yuri-scarbaci-lenio:
We also need this, and while `react-dnd` is good, we really prefeer using the simpler and leaner `react-sortable-hoc`

yuri-scarbaci-lenio:
Update the code to latest and Created a new PR implementing the same functionality as this PR with the latest code, in case you change your mind about how useful it could be :) 

https://github.com/clauderic/react-sortable-hoc/pull/749

PR #289, opened 2017-10-05
Add support for CSS grids

andrewcoelho:
Hey, first off thanks for the great project!

This PR addresses a small bug I ran into using the HOC's grid functionality with CSS grid styles.

Currently, items in a CSS grid are not translated correctly while sorting. This is because the translate offset is only using an item's `margin` to determine its translation amount. In a CSS grid, the `grid-gap`, `grid-column-gap`, and `grid-row-gap` properties can be used on a container to add spacing between its child elements. 

Also whipped up a repro of the issue, try dragging items around and you'll see the items being placed incorrectly during sorting and then shift to the correct positions on sort end.

https://codesandbox.io/s/vm917ww2z3

PR #487, opened 2019-01-19
Added CSS grid support

mmazzarolo:
Hey, cool library!
I just started using `react-sortable-hoc` and I noticed it doesn't work well with CSS grid (because of the grid gaps).     
In [Another Tab](https://github.com/mmazzarolo/chrome-another-tab) we're using CSS grid to show the bookmarks, so I investigated a little and I've found an open PR (#289) that adds the CSS grid support and, after updating it to the current state of the `master` branch, I tried it on the project and it has been working really well so far.  

Since the build process of this library is not running on post-install implementing it from a branch is not straight-forward, so I'm re-submitting the updated PR.  

All the work was done in #289 though, I just merely updated it (might need a bit more testing?). 

FYI here is a GIFs of the PR result on a CSS grid layout:

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -9,6 +9,7 @@ import {
   events,
   vendorPrefix,
   limit,
+  getContainerGridGap,
   getEdgeOffset,
   getElementMargin,
   getLockPixelOffset,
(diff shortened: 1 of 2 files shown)


lukegaluszka:
That looks great. Is there any option to get that feature in?

clauderic:
Closed in favour of https://github.com/clauderic/react-sortable-hoc/pull/657, thanks @mmazzarolo

PR #500, opened 2019-01-30
Fix virtualized collection grid bug

### LINKED ISSUE #412: Cannot read property 'left' of undefined with react-virtualized Grid
Integration with `react-virtualized` Grid seems to work pretty smooth for me.

However an error occurs when you drag and scroll an item. The nodes are being unmounted and there is no `prevNode` when scrolling down.

Changing this:
```js
translate.x = prevNode.edgeOffset.left - edgeOffset.left;
translate.y = prevNode.edgeOffset.top - edgeOffset.top;
```
https://github.com/clauderic/react-sortable-hoc/blob/master/src/SortableContainer/index.js#L637-L638

to this:
```js
if (prevNode) {
  translate.x = prevNode.edgeOffset.left - edgeOffset.left;
  translate.y = prevNode.edgeOffset.top - edgeOffset.top;
}
```

and for scrolling up look for `nextNode` and change it to this:
```js
if (nextNode) {
  translate.x = nextNode.edgeOffset.left - edgeOffset.left;
  translate.y = nextNode.edgeOffset.top - edgeOffset.top;
}
```

fixes the issue. Animation doesn't seem to be affected.

Maybe there is a better fix but at least `react-sortable-hoc` is working with `react-virtualized` Grid and +1000 items ✌️

I will dig further into this to figure out why the Grid is causing this issue and the List doesn't.

@clauderic if you have any ideas or can point me in the right direction that would be great.

Seems to be the same as this https://github.com/clauderic/react-sortable-hoc/pull/326.

If anyone wants to test it, use my fork:
https://github.com/superandrew213/react-sortable-hoc/tree/fix-react-virtualized-grid-with-dist

hubdrpl:
Fixes #412

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -671,8 +671,10 @@ export default function sortableContainer(
                 // If it moves passed the right bounds, then animate it to the first position of the next row.
                 // We just use the offset of the next node to calculate where to move, because that node's original position
                 // is exactly where we want to go
-                translate.x = nextNode.edgeOffset.left - edgeOffset.left;
-                translate.y = nextNode.edgeOffset.top - edgeOffset.top;
+                if (nextNode) {
+                  translate.x = nextNode.edgeOffset.left - edgeOffset.left;
+                  translate.y = nextNode.edgeOffset.top - edgeOffset.top;
+                }
               }
               if (this.newIndex === null) {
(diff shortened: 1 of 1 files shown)


clauderic:
Released in 1.54, thanks!

PR #501, opened 2019-02-05
Keyboard sorting

### LINKED ISSUE #77: Feature Request: Keyboard navigation
First of all, thanks for this amazing helper library to support drag and drop within list.

The feature request: Be able to drag and drop within list by using the keyboard.
This can be implemented using by a combination of the **Shift + Arrow keys** or **CMD / CTRL + Arrow keys**.

In my mind there are 2 possible behaviours for this feature. 
- On key press the active item will move the list item every 1 ms 1px further.
- On key down the active item will move to the next position in the given direction.

I would be happy to help / or implement this feature.

b-ryu:
Resolves #77 

This PR introduces keyboard-accessible sorting. Users are now able to sort items using only their keyboard.

Devs can set `tabIndex` on their `SortableElement`s. Once focused:
- **Space** picks up an item
- **Arrow UP/LEFT** moves the picked-up item backwards one place in the list
- **Arrow DOWN/RIGHT** moves the picked-up item forwards one place in the list
- **Space** again to drop the item
- **ESC** to cancel a sort

**Known issue(s)**
On `react-infinite`, occasionally (every couple hundred items) it'll stop sorting prematurely, due to what I believe is the virtualized DOM not appending items to the list fast enough for the sorting to continue. <s>The fix is simply to stop keyboard sorting, scroll down a bit with the mouse (or tab forward) and continue. Currently looking into possible solutions.</s> The simplest fix that I found was just to increase the preload values passed to `Infinite`'s props, specifically `preloadBatchSize` and `preloadAdditionalHeight`.


src/SortableContainer/index.js
```diff
@@ -851,6 +864,152 @@ export default function sortableContainer(
       );
     }
 
+    handleKeyDown = (event) => {
+      const {key} = event;
```
clauderic: Using `event.key` has inconsistent browser support, for instance, `ArrowLeft` is `Left` in IE/Edge, prefer using `event.keyCode`

src/SortableContainer/index.js
```diff
@@ -851,6 +864,152 @@ export default function sortableContainer(
...
+          break;
+        case 'ArrowDown':
+        case 'ArrowRight':
+          this.keyMove(1, event);
+          break;
+        case 'ArrowUp':
+        case 'ArrowLeft':
+          this.keyMove(-1, event);
+          break;
+      }
```
clauderic: I think you should consider having the escape key to cancel the operation

src/SortableContainer/index.js
```diff
@@ -851,6 +864,152 @@ export default function sortableContainer(
...
+    handleMouseDownAnywhere = () => {
+      if (this.manager.activeKeyLift) {
+        this.keyDrop();
+      }
+    };
+
+    // To deal with blurring on rerender
+    handleKeyLiftedElementBlur = (event) => {
+      event.target.focus();
+    };
```
clauderic: Not sure I fully understand what this is for, would you mind elaborating a bit?

src/SortableContainer/index.js
```diff
@@ -851,6 +864,152 @@ export default function sortableContainer(
...
+        this.manager.refs[collection].length - 1,
+      );
+
+      if (newIndex !== newerIndex) {
+        const swapNode = this.manager.getOrderedRefs(collection)[newerIndex].node;
+        if (newIndex < newerIndex) {
+          node.parentNode.insertBefore(swapNode, node);
+        } else {
+          swapNode.parentNode.insertBefore(node, swapNode);
+        }
```
clauderic: I'm not sure how I feel about this approach, I would prefer using `transforms` to move the elements into place as is done everywhere else in the library while sorting. The logic would be a bit more verbose, but the overall user experience would be more aligned with the mouse/touch experience.

src/SortableContainer/index.js
```diff
@@ -851,6 +864,152 @@ export default function sortableContainer(
...
+        this.keyAutoScroll();
+
+        swapNode.sortableInfo.index = newIndex;
+        node.sortableInfo.index = newerIndex;
+        this.manager.activeKeyLift.newIndex = newerIndex;
+      }
+
+      if (typeof onSortMove === 'function') {
+        onSortMove(event);
+      }
```
clauderic: I haven't spent too much time thinking about this yet, but I think it might be a bit confusing for consumers of the library to have two different event types for `onSortMove`

clauderic (commented):
Hey @b-ryu,

Thanks for taking the time to submit this PR! Left a few comments / questions / suggestions, I think this is going in the right direction, my main concern is with the way the elements are moved in the DOM currently, I'd prefer using transforms in order to offer the possibility to animate the elements as they shift positions.

On a separate note, I think it'd also be nice from an accessibility perspective to announce what is going on to assist users with visual (or other) impairments: 

This doesn't have to happen in this PR, but something worth considering :)

src/SortableContainer/index.js
```diff
@@ -851,6 +864,152 @@ export default function sortableContainer(
       );
     }
 
+    handleKeyDown = (event) => {
+      const {key} = event;
+      const ignoreKeyDown = this.manager.activeKeyLift == null && key !== ' ';
+
+      if (ignoreKeyDown) {
```
clauderic: You should also early return while a touch / mouse sort is in progress, otherwise it can lead to both the mouse and keyboard operations to happen concurrently.
 
I also think we should `preventDefault` when the tab key is pressed while dragging (both for keyboard and mouse/touch sorting)

src/SortableContainer/index.js
b-ryu: Can do 👍

src/SortableContainer/index.js
b-ryu: Noted, thanks! 👍

src/SortableContainer/index.js
b-ryu: I ran into some issues where the sorted element was losing focus while been moved through the list. This is mainly a by-product of the DOM manipulation strategy though, I'll see if I can get rid of this in my next push.

src/SortableContainer/index.js
b-ryu: Right, makes sense. I actually starting trying out different approaches when I found out this one didn't work well with virtualization libraries 😄 I'll try fleshing out a different implementation and putting it up.

src/SortableContainer/index.js
b-ryu: Makes sense 👍

src/SortableContainer/index.js
b-ryu: Right, I'll leave it out 👍

natew:
Just want to say it's super encouraging to see signs of life with this library. Thanks @clauderic!

I want to submit a PR that just moves to the new Context API (see: https://twitter.com/dan_abramov/status/1097866569701621763). I imagine it's a small PR but was worried it may not be accepted ever. Will try now!

src/SortableContainer/index.js
```diff
@@ -160,19 +166,22 @@ export default function sortableContainer(
             );
           }
         }
+        this.container.addEventListener('keydown', this.handleKeyDown);
       });
     }
 
     componentWillUnmount() {
```
clauderic: Need to make sure to clean up the `this.helper` node from the DOM if it exists here, otherwise it will stay appended to the DOM when this component unmounts.

Now that I think of it, this was an issue as well before, though quite unlikely to happen when sorting with mouse / touch input. It's much more likely to happen with keyboard sorting if the user forgets to drop the item.

src/SortableContainer/index.js
```diff
@@ -890,6 +947,272 @@ export default function sortableContainer(
...
+
+      switch (keyCode) {
+        case KEYCODE.SPACE:
+          if (this.manager.activeKeyLift) {
+            this.keyDrop(event);
+          } else {
+            this.keyLift(event);
+          }
+          break;
+        case KEYCODE.DOWN:
```
clauderic: This is probably a stretch for this PR, so let's keep it for a subsequent PR, but from a UX perspective, arrow up / down in a grid setup should probably move to the previous / next row

src/SortableContainer/index.js
```diff
@@ -289,8 +298,8 @@ export default function sortableContainer(
       }
     };
 
-    handlePress = async (event) => {
-      const active = this.manager.getActive();
+    handlePress = async (event, keyboard = false) => {
+      const active = keyboard ? this.manager.getActiveKeyLift() : this.manager.getActive();
```
clauderic: We should give focus to the helper node while sorting with keyboard

src/.stories/index.js
```diff
@@ -42,7 +44,7 @@ const Item = SortableElement((props) => {
       }}
     >
       {props.shouldUseDragHandle && <Handle />}
-      <div className={style.wrapper}>
+      <div className={style.wrapper} tabIndex={tabIndex}>
```
clauderic: I'm a bit curious, why make only the span tabbable instead of the entire item?

For components using a drag handle, we should probably document that only the handle should be made tabbable. For now, would you mind updating the storybook for drag handle so that only the handle is tabbable?

src/SortableContainer/index.js
```diff
@@ -148,6 +151,9 @@ export default function sortableContainer(
         this.scrollContainer = useWindowAsScrollContainer
           ? this.document.scrollingElement || this.document.documentElement
           : this.container;
+        this.keyScrollContainer = useWindowAsScrollContainer
```
clauderic: I'm not sure I follow why this is needed, could you clarify the rationale for having a `keyScrollContainer`?

src/utils.js
```diff
@@ -137,6 +137,32 @@ export function getEdgeOffset(node, parent, offset = {top: 0, left: 0}) {
...
+  } catch (err) {
+    return null;
+  }
+
+  if (overflow === 'auto' || overflow === 'scroll') {
+    return el;
+  } else {
+    return getScrollingElement(el.parentNode);
+  }
+}
```
clauderic: This is a nice addition, though sort of unrelated to the current PR as far as I can tell? Unless it's needed specifically for this PR, I'd prefer introducing it in a separate PR

clauderic (commented):
Hey @b-ryu,

This is looking great, tophatted a bunch of different scenarios and it works surprisingly well. You mentioned there were issues with `react-infinite`, have you been able to replicate similar issues with the `react-virtualized` or `react-tiny-virtual-list` examples?

I've left a few questions / suggestions for now. I'll take a closer look at the logic for moving the items around over the weekend. 

As an aside, I'm a bit worried about the amount of complexity and alternative code paths it introduces to the codebase, but that's probably more of a symptom of how the library currently works. In the future, I'd like to refactor the translations to be based on deterministic state updates and only have one code path for translates using `getBoundingClientRect` to swap items around.

src/SortableContainer/index.js
b-ryu: Will do 👍

src/.stories/index.js
b-ryu: Makes sense 👍

src/SortableContainer/index.js
b-ryu: I was having issues specifically with the autoscroll in the nested list example. Moving an item within one of the categories wouldn't trigger the autoscroll since `this.scrollContainer` pointed to the category, not the larger scrolling list itself. 

This isn't really an issue with the mouse scrolling, but for keyboard scrolling you'd have the helper spilling over the outside of the container. I wasn't sure whether to change the initialization of `this.scrollContainer` and possibly change the existing mouse sorting implementation, or to add a separate pointer for my own purposes, so I opted for what I thought was the safer route until I could get your thoughts on the matter 😅

b-ryu:
First of all, I want to thank you for taking the time to review this PR 😄👍 

> You mentioned there were issues with `react-infinite`, have you been able to replicate similar issues with the `react-virtualized` or `react-tiny-virtual-list` examples?

Not that I know of, the issue I mentioned seems only to arise when reaching the 300th or so item in a `react-infinite` list. I seem to be able to shuffle through the entire length of a `react-virtualized` and `react-tiny-virtual-list` without any issues, but that may take some more scrutiny to confirm.

> As an aside, I'm a bit worried about the amount of complexity and alternative code paths it introduces to the codebase

Good point 👍 I initially considered inserting the logic for some of the key sorting methods into the existing methods since most of the underlying structure was the same, but found that it was a little messy, so I opted to share the ones I could use with minimal modification and extract the rest into their own set of methods.

> In the future, I'd like to refactor the translations to be based on deterministic state updates and only have one code path for translates using `getBoundingClientRect` to swap items around.

 Makes sense, caching the initial `getBoundingClientRect` values would really simplify the positioning logic, for `keyUpdatePosition` at least 😄

src/SortableContainer/index.js
clauderic: I think this is probably just a symptom of `this.scrollContainer` default not being smart enough. Perhaps this should be improved in a separate PR  first along with https://github.com/clauderic/react-sortable-hoc/pull/501/files#r263577386 so as to only have one reference to `scrollContainer`.

src/SortableContainer/index.js
```diff
@@ -160,19 +166,25 @@ export default function sortableContainer(
...
-      if (this.container) {
-        for (const key in this.events) {
-          if (this.events.hasOwnProperty(key)) {
-            events[key].forEach((eventName) =>
-              this.container.removeEventListener(eventName, this.events[key]),
-            );
-          }
+      if (this.helper && this.helper.parentNode) {
+        this.helper.parentNode.removeChild(this.helper);
+      }
```
clauderic: LGTM 👍

src/SortableContainer/index.js
```diff
@@ -890,6 +954,272 @@ export default function sortableContainer(
...
     }
 
+    handleKeyDown = (event) => {
+      const {keyCode} = event;
+      const {shouldCancelStart} = this.props;
+
+      if (
+        shouldCancelStart(event) ||
+        this.manager.active ||
+        (this.manager.activeKeyLift == null && (
```
clauderic: Do we really need `activeKeyLift` (as opposed to also using `active` for keyboard sorting)?

src/SortableContainer/index.js
```diff
@@ -890,6 +954,272 @@ export default function sortableContainer(
...
+          keyCode !== KEYCODE.SPACE ||
+          !closest(event.target, (el) => el.sortableInfo != null)
```
clauderic: I don't think this is specific enough. If there are any other focusable elements within the sortable container, pressing space while they are focused will also trigger sorting. We should only treat the event as valid if the `event.target` is a SortableHandle or a SortableElement (`el.sortableInfo != null`) and there is no SortableHandle within. In the event that there is a SortableHandle within a SortableElement, only the SortableHandle element having focus should be treated as a valid sort event, even if the SortableContainer is focusable / focused.

I think you should consider adding a stress test example with many common use cases such as inputs within sortable elements to the Storybook specifically for keyboard sorting.

beefchimi:
Just tested this out locally - awesome work! Really excited about this. I didn't look too closely at the code - my only comments though would have been nitpicky style changes for the examples 🙃 nothing to worry about!

src/SortableContainer/index.js
b-ryu: I'll try and get rid of it. I think my initial reasoning for having it was to keep the keyboard sorting logic from stepping on the toes of the original implementation, but having just one `active` makes more sense 👍

src/SortableContainer/index.js
b-ryu: Will do 👍

src/SortableContainer/index.js
b-ryu: Follow up: made a PR for this subissue #507

src/SortableContainer/index.js
```diff
@@ -394,21 +412,23 @@ export default function sortableContainer(
...
         }
 
-        if (keyboard) {
-          // Adjust the nodes if the helper changes size
-          this.keyAnimateNodes(true);
-        }
+        // if (keyboard) {
+        //   // Adjust the nodes if the helper changes size
+        //   this.keyAnimateNodes(true);
+        // }
```
clauderic: Not sure why this was needed, commented it out for the time being

src/SortableContainer/index.js
```diff
@@ -725,11 +761,17 @@ export default function sortableContainer(
...
       const {disableAutoscroll} = this.props;
+      const {isKeySorting} = this.manager.active;
 
       if (disableAutoscroll) {
         return;
       }
 
+      if (isKeySorting) {
+        // TBD
+        return;
```
clauderic: Still need to implement auto-scrolling for keyboard sorting

src/SortableContainer/index.js
```diff
@@ -813,215 +860,57 @@ export default function sortableContainer(
...
 
-      this.keyAnimateNodes();
+      const newNode = nodes[newIndex].node;
+      const edgeOffset =
+        newNode.edgeOffset || getEdgeOffset(newNode, this.container);
+
+      this.handleSortMove({
+        pageX: this.containerBoundingRect.left + edgeOffset.left,
+        pageY: this.containerBoundingRect.top + edgeOffset.top,
+      });
```
clauderic: We create a mock event here with the coordinates of where the mouse would be if this were a mouse event, so as to be able to re-use the logic for animating nodes. While the copy paste approach worked, it would have made maintaining and contributing to this library exponentially more challenging

src/SortableContainer/index.js
b-ryu: I was running into issues with correctly positioning the helper and the elements in the cases where there's resizing during sorting.

src/SortableContainer/index.js
```diff
@@ -708,6 +860,121 @@ export default function sortableContainer(
       );
     }
 
+    handleKeyDown = (event) => {
+      const {keyCode, target} = event;
+      const {shouldCancelStart, useDragHandle} = this.props;
+
+      const isValidSortingTarget = useDragHandle
+        ? isSortableHandle(target)
+        : target.sortableInfo;
```
clauderic: This check also needs to verify that the `SortableElement` is not disabled, otherwise disabled items can be sorted

src/SortableContainer/index.js
```diff
@@ -517,6 +626,10 @@ export default function sortableContainer(
         if (!edgeOffset) {
           edgeOffset = getEdgeOffset(node, this.container);
           nodes[i].edgeOffset = edgeOffset;
+          // While we're at it, cache the boundingClientRect, used during keyboard sorting
+          nodes[i].boundingClientRect = getScrollAdjustedBoundingClientRect(
+            node, containerScrollDelta,
+          );
```
clauderic: I'm somewhat worried about the performance overhead of getting the boundingClientRect of each node, could we limit this only to keyboard sorting for the time being?

src/SortableContainer/index.js
```diff
@@ -708,6 +860,121 @@ export default function sortableContainer(
...
+      const {target} = event;
+      const node = closest(target, (el) => el.sortableInfo != null);
+      const {index, collection} = node.sortableInfo;
+
+      this.initialFocusedNode = target;
+
+      this.manager.active = {
+        index,
+        collection,
+        isKeySorting: true,
```
clauderic: I think it might make more sense to keep track of `isKeySorting` directly on the instance of `this.manager` instead of `this.manager.active`

clauderic (commented):
Hey @b-ryu, great work on the refactor, this is looking really close to being ready to merge 👌 

Gave this a fairly thorough tophat, it works great. There was only one small bug that I found with keyboard sorting of horizontal lists, it looks like there's an issue with the last index of the list, here's a gif demonstrating the issue:
Kapture 2019-04-20 at 19 42 12

I believe the same issue exists with mouse sorting actually, so might not be a huge deal for the time being.

lowewenzel:
I am having trouble implementing this, and also hard to find examples on where `tabIndex={0}` should be. I am using this with `react-virtualized`. I tried the `tabIndex` on the `SortableElement` as well as the outermost `<div>` within the `SortableElement` implementation.

Sorry for bringing up this PR, but docs are pretty minimal on accessibility support. Thanks in advance!

src/SortableContainer/index.js
andersahp: @clauderic is there plans to include this functionality for up/down soon? I mean down arrow means swap element with element below in grid and similar for up arrow. Seems this is not possible at the moment

vaynevayne:
It won't work at all, emmm...

```jsx
const GridItem = SortableElement(({ value }) => (
  <div style={gridItemStyles} tabIndex={0}>
    {value}
  </div>
))
const Grid = SortableContainer(({ items }) => (
  <div style={gridStyles}>
    {items.map((col, index) => (
      <GridItem key={col.key} index={index} value={col.title} tabIndex={0} />
    ))}
  </div>
))
```

PR #507, opened 2019-03-12
Improve scroll container behaviour

b-ryu:
[Subissue of #501](https://github.com/clauderic/react-sortable-hoc/pull/501#discussion_r264287089)

This PR improves the scroll container default by checking the element's ancestors for appropriate scrolling styles. It also replaces references to `this.container` with those to `this.scrollContainer` in the autoscrolling logic.

This mainly impacts the nested list use case; other use cases (including window as scroll container) should remain unaffected.

If this gets merged in, I'll put up a commit in #501 to consolidate the changes.

_Sorry for the terrible GIF quality_

| Before | After |
|---|---|
| masterscroll | sortingscroll

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -174,24 +174,22 @@ export function getLockPixelOffset({lockOffset, width, height}) {
   };
 }
 
-export function getScrollingParent(el) {
-  try {
-    const {
-      overflow,
-      overflowX,
-      overflowY,
-    } = window.getComputedStyle(el);
+function isScrollable(el) {
+  const computedStyle = window.getComputedStyle(el);
+  const overflowRegex = /(auto|scroll)/;
+  const properties = ['overflow', 'overflowX', 'overflowY'];
 
-    if (
-      overflow === 'auto' || overflow === 'scroll' ||
-      overflowX === 'auto' || overflowX === 'scroll' ||
-      overflowY === 'auto' || overflowY === 'scroll'
-    ) {
-      return el;
-    } else {
-      return getScrollingParent(el.parentNode);
-    }
-  } catch (err) {
+  return properties.find((property) =>
+    overflowRegex.test(computedStyle[property]),
+  );
+}
+
+export function getScrollingParent(el) {
+  if (!el) {
     return null;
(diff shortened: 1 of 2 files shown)


src/utils.js
```diff
@@ -173,3 +173,18 @@ export function getLockPixelOffset({lockOffset, width, height}) {
...
+
+export function getScrollingParent(el) {
+  let overflow;
+  try {
+    overflow = window.getComputedStyle(el).overflow;
+  } catch (err) {
+    return null;
+  }
+
+  if (overflow === 'auto' || overflow === 'scroll') {
```
clauderic: I think you also need to account for `overflow-x` and `overflow-y`

src/utils.js
```diff
@@ -173,3 +173,25 @@ export function getLockPixelOffset({lockOffset, width, height}) {
...
 }
+
+export function getScrollingParent(el) {
+  try {
+    const {
+      overflow,
+      overflowX,
+      overflowY,
+    } = window.getComputedStyle(el);
+
```
clauderic: What do you think of something along these lines:

```js
function isScrollable(el) {
  const computedStyle = window.getComputedStyle(el);
  const overflowRegex = /(auto|scroll)/;
  const properties = ['overflow', 'overflowX', 'overflowY'];
  
  return properties.find((property) =>
    computedStyle[property].matches(overflowRegex)
  );
}

function getScrollingParent(el) {
  if (isScrollable(el)) { 
  ...
}
```

src/utils.js
```diff
@@ -173,3 +173,23 @@ export function getLockPixelOffset({lockOffset, width, height}) {
...
+  const overflowRegex = /(auto|scroll)/;
+  const properties = ['overflow', 'overflowX', 'overflowY'];
+
+  return properties.find((property) =>
+    overflowRegex.test(computedStyle[property]),
+  );
+}
+
+export function getScrollingParent(el) {
+  if (!el) {
```
clauderic: This might need to be `!(el instanceof HTMLElement)`

PR #512, opened 2019-03-18
fixed cloning canvas context of dragged items

mohsom:
Hello, after installing the last release of the package I've encountered problem with sorting canvas items. After some debugging I found out that the problem is in copying canvas context on sort (it gets elements from array of items with the wrong index).


clauderic:
Published this fix in `1.8.1`

PR #526, opened 2019-04-16
Fix IE11 "Possible Unhandled Promise Rejection" error

### LINKED ISSUE #525: IE11 "Possible Unhandled Promise Rejection" error
Hi,

'react-sortable-hoc' version greater than 1.0.0 is not working on IE11 and throwing "Possible Unhandled Promise Rejection" console message.

Is this component no longer supporting IE11 on newer versions?

Thanks for any help.

ApacheEx:
Fixes #525

`Array.find()` doesn't work in [IE11](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) at all. 

So, this is super light-weight fix which allows supporting IE9+ browsers.

### DIFF
--- a/src/Manager/index.js
+++ b/src/Manager/index.js
@@ -22,10 +22,16 @@ export default class Manager {
   }
 
   getActive() {
-    return this.refs[this.active.collection].find(
+    const ref = this.refs[this.active.collection];
+
+    for (let i = 0; i < ref.length; i++) {
+      const { node } = ref[i];
+
       // eslint-disable-next-line eqeqeq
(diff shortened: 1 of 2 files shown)


clauderic:
You should include your own polyfills for these missing methods in the browsers that you wish to support.

ApacheEx:
I understand, but it's really small changes to avoid including polyfills. A lot of users do not like or have possibility to include polyfills.

ApacheEx:
If you don't wont / like to merge it. Could you add minimal supported version to README or wherever?

PR #527, opened 2019-04-16
Use vanilla loop to get rid of Symbol.iterator and Array.from dependencies

ApacheEx:
One more PR to support IE11 and avoid this issue:
image


clauderic:
You should include your own polyfills for these missing methods in the browsers that you wish to support.

PR #531, opened 2019-04-23
Pass isKeySorting to other handler props

b-ryu:
Passes `isKeySorting` to `updateBeforeSortStart`'s `SortStart` parameter.

Forgot to include this in #501 😅


clauderic:
Good catch! Same with `onSortOver` actually now that I think of it.

Would you mind also updating the function signature in the docs? 😄

clauderic:
Released with version `1.9.1`

PR #536, opened 2019-05-05
The helperContainer callback should provide the this context

Tofandel:
If I want to append the sorted element within the element that is being sorted (to not lose styling because there is a lot of css rules)
I need to pass helperContainer: () => { return this.container } as a prop, but the this is not passed as a context to the function so it doesn't work, adding .call(this) effectively addresses the issue

I was also wondering why append the element to the body and not to the container by default? It would prevent a lot of issues

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -1036,10 +1036,10 @@ export default function sortableContainer(
       const {helperContainer} = this.props;
 
       if (typeof helperContainer === 'function') {
-        return helperContainer();
+        return helperContainer.call(this);
       }
 
-      return this.props.helperContainer || this.document.body;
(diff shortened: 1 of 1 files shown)


pjshy:
I think the helperContainer callback provide the cloned sortable helper is useful sometime. But the context is complex.

Tofandel:
@clauderic Any motion/thoughts on the PR please? It's been pending for quite some time

Robin-Hoodie:
> I was also wondering why append the element to the body and not to the container by default? 

I was wondering the same thing. I imagine a lot of consumers of this library will need to work around this for the reasons you mentioned. Why not attach it to a more sensible default element or at least provide a simple option to attach it to the sortable container element?

PR #541, opened 2019-05-14
Remove browser field with umd bundle

TrySound:
`browser` field has a priority over `module` field in all bundlers. Adding it to package.json drops all goodness of esm bundle.
It should be used in a bit different way to support both node and browser fields.
https://github.com/uber/react-map-gl/blob/1da1398c063ffc3b8464744bde6781cb5dc3866e/package.json#L20-L24


TrySound:
Friendly ping @clauderic

TrySound:
Hi @clauderic.If you have some time could you publish latest changes?

Issue #118, opened 2017-01-13
Link inside of SortableElement breaks sort in FF


Moerphy:
Testcase: https://jsfiddle.net/tz2daph1/1/
When clicking on the list item it works correctly, when clicking the link itself it only moves a short distance with the mouse, then stops moving until mouseup. And then it follows the mouse again like it would normally until the next mouseup event.

This only happens in Firefox 50 for me, works in all other major browsers.

Moerphy:
Thanks for the fast fix! Seems like the bug still happens if the SortableContainer is using the distance property. I think moving the fix to handleStart instead of handlePress should fix that too.

https://jsfiddle.net/tz2daph1/2/

clauderic:
Riiiiiight! Will fix shortly, sorry about that!

clauderic:
Just pushed a fix to npm with `0.4.7`, give it a spin and let me know if all is well!

Moerphy:
Thanks, looks like it works perfectly now!

PR #544, opened 2019-05-22
Fix firefox event issue related to Button tag

benadam11:
This is the same issue as https://github.com/clauderic/react-sortable-hoc/issues/118 but it also applies to button on FF

Example of issue: https://84r5vjm2yl.codesandbox.io/ (FF only)

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -153,7 +153,11 @@ export default function sortableContainer(
          * prevent subsequent 'mousemove' events from being fired
          * (see https://github.com/clauderic/react-sortable-hoc/issues/118)
          */
-        if (!isTouchEvent(event) && [NodeType.Anchor, NodeType.Button].includes(event.target.tagName)) {
+        if (
+          !isTouchEvent(event) &&
+          (event.target.tagName === NodeType.Anchor ||
+            event.target.tagName === NodeType.Button)
+        ) {
           event.preventDefault();
         }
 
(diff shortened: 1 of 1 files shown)


src/SortableContainer/index.js
```diff
@@ -153,7 +153,7 @@ export default function sortableContainer(
          * prevent subsequent 'mousemove' events from being fired
          * (see https://github.com/clauderic/react-sortable-hoc/issues/118)
          */
-        if (!isTouchEvent(event) && event.target.tagName === NodeType.Anchor) {
+        if (!isTouchEvent(event) && [NodeType.Anchor, NodeType.Button].includes(event.target.tagName)) {
```
clauderic:
```suggestion
        if (
          !isTouchEvent(event) &&
          (event.target.tagName === NodeType.Anchor ||
            event.target.tagName === NodeType.Button)
        ) {
```

clauderic: As much as I prefer the `.includes` sintax, let's avoid using it as this would be the first occurence of it in the lib and I often receive complaints from people unwilling to include polyfills in their app.

clauderic:
Actually upon further investigation I actually don't think this is the same issue as #118. I believe the issue you're pointing out isn't actually specific to Firefox, and rather is just the default implementation for `shouldCancelStart`, see the default implementation here: https://github.com/clauderic/react-sortable-hoc/blob/master/src/SortableContainer/defaultShouldCancelStart.js#L10

Most people want their buttons to still be interactive and not cause a sort event to begin, so by default `react-sortable-hoc` prevents sort events from being triggered on button elements.

I'm going to be reverting this PR

PR #573, opened 2019-07-18
In Grids - Grid overflow fixed when items moved from backward.(Wrapper changed from window to custom container as an optional prop)

nilavrusty:


knarkunas:
Is this related to it? Could provide more text in description and less in title.
https://github.com/clauderic/react-sortable-hoc/issues/577

Issue #510, opened 2019-03-16
Combining items


giacomocerquone:
Is something like this: https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/guides/combining.md possible with this lib?
I need grid and merge support (for merge I mean exactly dragging one item to drop it over another item)

I think this would be simple to implement. I'd just need a "delayAnimationBasedOnDistance" prop (clearly the name isn't too good) in order to not start the animation if the dragged items moves only of (let's say) 200px and do something on over, if left there the "onSortEnd" function will take care of remove that item from the array and do something else with it.

I believe this "delayAnimationBasedOnDistance" could also be implemented without touching the codebase, basically I could register an onMove callback and keep the item disabled (with the respective prop) until the cursor moved of N pixels.

What do you think? Should this work for you? If so I could do a PR.

clauderic:
How does this differ from the `distance` prop? 

With regards to combining items, it should be feasible assuming you manage the selection / combining of the items in your app and leverage the `updateBeforeSortStart` prop

giacomocerquone:
@clauderic it differs from the distance prop since when you set that prop, if you don't do a certain distance with the cursor while clicking, it will not enable the "animation" of the item being removed from the flow. What I ask is that if you are already dragging around an element and don't do a certain distance, THE OTHER items (so not the one you're already dragging) will not animate.

Honestly I don't get how I could use in this scenario the updateBeforeSortStart... the problem is that if I place an item over another item, the one below will very soon move at its left or right to make room for the item I'm dragging

giacomocerquone:
It this is not enough I can make a gif

clauderic:
If you could provide a gif, that would be helpful, as I'm still not sure I understand your use-case completely.

I just put together an example for grouping items that I'm going to add to the storybook:

<img src="https://user-images.githubusercontent.com/1416436/54495093-437d2200-48b7-11e9-885b-05e64c554615.gif" width="400" />

See https://github.com/clauderic/react-sortable-hoc/pull/511 for details and the code, I think this is similar to what you're describing and could be a good starting point for you to achieve what you're looking for

giacomocerquone:
Sorry but it's true that I said I could do a gif, but this example that I previously linked I believe is pretty clear, even more now compared to the gif you posted. Look at the differences between these two:

[...]

PR #598, opened 2019-08-28
Added dropover functionality

leanderlee:
Provides a solution for issues like #510, looking to detect drop events.

Adds `onDropOver` and `onDropCancel` functions that get called when dragging over a new index and when no longer dropping over any indices, respectively.

Currently only works on the y axis, but just wanted to do a PR to see what people think of this direction.


hardik98:
Hey @leanderlee is there any working example out there for this  feature ?

Issue #333, opened 2018-01-16
Reorders items of different sizes in wrong order


jacobp100:
I have a list of a large item, a small item, and another large item. When I drag the first item down, the third item is moved before the second. See below for a video demonstrating this. I have the options,

```jsx
<MyComponent
  ...
  distance={8}
  lockAxis="y"
  lockToContainerEdges
/>
```

![ezgif-3-ec88f6e977](https://user-images.githubusercontent.com/7275322/34986057-0d3d6b1c-faae-11e7-9e72-b3731c650ea4.gif)

jordanbb:
Also running into the same problem but with a horizontal list. If you drag a large list item from left to right in between a small list item and another large list item, it will "skip" the space between small and right large list item and move everything over to the left. Move a large list item from right to left appears to work fine though. I would show an example but @jacobp100 example illustrates it well, but just vertical.

ncolboc:
I encountered the same problem as @jordanbb mentions with horizontal list. react-sortable-hoc seems to have limited support with items which have not same size.

PR #602, opened 2019-08-29
shift forward/backward when the dragging element's midline crosses a …

evanc:
Hello,

This resolves the issue reported in #428 and #333 - when the height difference between rows is extreme, they sort out-of-order.

[This sandbox demonstrates the issue](https://codesandbox.io/s/react-sortable-hoc-starter-pf3be?fontsize=14). Try dragging *Item 1* between *Item 2* and *Item 3*.

Here's what happens:
Screen Recording 2019-08-29 at 03 52 PM

Item 3 slides up before Item 2.

Here's the updated behavior:
https://codesandbox.io/s/react-sortable-hoc-starter-vjllg

Screen Recording 2019-08-29 at 04 42 PM

I changed the logic that determines when to shift a list item. In the released version of react-sortable-hoc, the list item shifts differently depending on whether you're moving a taller item over a shorter item (or wider/narrower as it happens). When items are of greatly varying size, they shift out of order.

My change will cause the list item being dragged over to shift whenever the midpoint of the dragged item crosses the midpoint of the list item. This provides a much more consistent experience.

I realize this is a fairly noticeable change in some cases, so I've also prepared [an updated version of the storybook](http://evanc-react-sortable-hoc-pr.surge.sh/?path=/story/general-layout-vertical-list--variable-heights) so you can see how this change affects other types of sortable lists. I think it feels either the same or better in every case.


evanc:
actually this seems like it screws up keyboard sorting. I'll update this with a fix.

Mathster0598:
Would love to see this PR live, been having the same problems.

Luncher:
I also encountered the same problem. How did you solve it later?

@evanc @Mathster0598 @nickdekesel

Issue #460, opened 2018-12-21
react-sortable-hoc broken in ConcurrentMode


Kovensky:
I haven't really made a reproduction case yet, but just enabling `ConcurrentMode` in my tree that includes `react-sortable-hoc` is sufficient to break it as it is used in my application.

Related to https://github.com/clauderic/react-sortable-hoc/issues/444.

When `ConcurrentMode` is enabled, the helpers that show up when dragging starts never disappear, and the SortableList also does not correctly update when a reorder does happen.

clauderic:
Hey @Kovensky,

That's no good, I haven't tried using `react-sortable-hoc` with `ConcurrentMode` yet, but I will investigate. 

You mentioned #444, what makes you think the issues you noticed are related to that issue?

Kovensky:
`SortableElement` does side effects in `componentWillReceiveProps`, which is exactly the kind of thing that is expected to break in `ConcurrentMode` and why it is a `StrictMode` warning. I believe what is happening is a race condition between `SortableElement` removing itself from the `Manager` and `SortableContainer`'s `handleSortEnd` logic being invoked; as `SortableElement#componentWillReceiveProps` might be invoked multiple times without committing to the DOM as part of a single yieldy render.

clauderic:
Right. I have a PR open that updates the behaviour of the `disabled` prop, and also changes from using `componentWillReceiveProps` to `componentDidUpdate` in `SortableElement` to manage the side-effects. I haven't tested it with ConcurrentMode yet, but I'm expecting it should work properly.

Kovensky:
I haven't tested it but it does look more correct 👍 (fixed link: https://github.com/clauderic/react-sortable-hoc/pull/472)

As for the remaining warnings:

- String ref can probably easily be fixed by using a function ref (or `createRef`, but if you do that you might as well do the legacy context migration at the same time).
- Legacy context is more of a performance / correctness issue but I don't think it'll break in `ConcurrentMode`, as long as context updates only happen in commit lifecycles (`componentDidMount` / `componentDidUpdate` / `componentWillUnmount`). AFAICT context updates don't really happen at all (`createContext`-wise); the `manager` is created once and is only internally mutated.
- `findDOMNode` is deprecated because it only returns the first DOM node for components that produce more than one, but it otherwise should be thread-safe to call in `componentDidMount` / `componentDidUpdate`. It's still bad to call it in other lifecycles, in render, or in event handlers.

PR #603, opened 2019-08-30
Use new context type to prevent de-opts in React

natew:
Using the new context type should fix performance and be forward-comparable with Concurrent mode.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@o/react-sortable-hoc",
-  "version": "1.10.2",
+  "version": "1.10.4",
   "description": "Set of higher-order components to turn any list into a sortable, touch-friendly, animated list",
   "author": {
     "name": "Clauderic Demers",
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "react-sortable-hoc",
+  "name": "@o/react-sortable-hoc",
   "version": "1.10.1",
   "description": "Set of higher-order components to turn any list into a sortable, touch-friendly, animated list",
   "author": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@o/react-sortable-hoc",
-  "version": "1.10.1",
+  "version": "1.10.2",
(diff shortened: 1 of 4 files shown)


natew:
Updated it to also handle the createRef() issue, see: https://github.com/clauderic/react-sortable-hoc/issues/460

Only thing left would be findDOMNode, but that seems to be fine for now.

package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "react-sortable-hoc",
-  "version": "1.10.1",
+  "name": "@o/react-sortable-hoc",
+  "version": "1.10.4",
```
clauderic: Would you mind removing this change from this PR?

src/SortableContainer/index.js
```diff
@@ -1029,10 +1024,16 @@ export default function sortableContainer(
...
 
+    wrappedInstance = React.createRef();
+
     render() {
-      const ref = config.withRef ? 'wrappedInstance' : null;
+      const ref = config.withRef ? this.wrappedInstance : null;
 
-      return <WrappedComponent ref={ref} {...omit(this.props, omittedProps)} />;
+      return (
+        <SortableContext.Provider value={{manager: this.manager}}>
```
clauderic: Would you mind moving the value of the provider outside of the render method as a class property? This way it won't re-create a new object on every single render.

src/SortableHandle/index.js
```diff
@@ -21,11 +21,13 @@ export default function sortableHandle(
...
       );
-      return this.refs.wrappedInstance;
+      return this.wrappedInstance.current;
     }
 
+    wrappedInstance = React.createRef();
+
     render() {
-      const ref = config.withRef ? 'wrappedInstance' : null;
+      const ref = config.withRef ? this.wrappedInstance : null;
```
clauderic: In the future I think we may actually want to use `React.forwardRef()`, but for now this is an okay change

natew:
Don't think I have time for this unfortunately, but will leave open for anyone else.

lindskogen:
I continued this task in https://github.com/clauderic/react-sortable-hoc/pull/624 🎉

PR #604, opened 2019-09-02
Clear autoscroller when autoscroll is disabled

superandrew213:
If `disableAutoscroll` is set to `true` by another event while user is dragging, scroll still continues.

Must also clear `autoScroller` when autoscroll gets disabled.


superandrew213:
Sure no problem.

PR #605, opened 2019-09-05
fix: helperClass transform not working

LouisMelo:
When `helperClass` props is given a `transform` style, it is not working during the drag process.

The transform in helperClass should be treated specially because it will be override by `setTranslate3d`.


LouisMelo:
@clauderic Hello there! Would you please help me review this little fix?

clauderic:
Hey @LouisMelo, could you share a CodeSandbox example where this fix would be useful?

LouisMelo:
@clauderic Sure, [here](https://codesandbox.io/embed/cool-satoshi-zm0ws) you can find where is the issue I want to fix~

LouisMelo:
@clauderic Hi, would you like have a second review on this PR please~

h-labushkina:
@clauderic Could you please proceed with this PR sounds like it solve the safari issue that I faced with

PR #635, opened 2019-11-24
fix: Ctrl click triggers sorting

### LINKED ISSUE #256: Right Click triggers sort
At least for me, even when trying to stopPropagation on the contextmenu event it starts dragging the element. Not sure if contextmenu clicks also trigger normal click, and if that is true, how to disable the second regular event from firing.

quanglam2807:
Resolve #256.

I also add `e.which === 3` condition as it is preferred by modern browser. Ref: https://api.jquery.com/event.which/, https://stackoverflow.com/a/2405835

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -126,7 +126,14 @@ export default function sortableContainer(
     handleStart = (event) => {
       const {distance, shouldCancelStart} = this.props;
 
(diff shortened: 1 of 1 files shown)


kytwb:
Looking forward to this one getting merged 🙏

adrianhelvik:
Is this repo still maintained?

PR #643, opened 2019-12-07
Fixed layout thrashing - performance issue

aseferov:
**Problem**:
There is huge performance problem while handling mouse move first time. It freeze 1-2s before moving when sortable element count increased. If sortable element has more complicated dom tree it getting even worse.

**What cause the problem**
While handling first mousemove there is for loop applying style `transform: translate3d(0,0,0)` for each node. On the next iteration of the loop, the browser has to account for the fact that styles have changed since offsetWidth was last requested (in the previous iteration), and so it must apply the style changes, and run layout. This will happen on every single iteration!.

slower

**Solution**
Moving applying style outside of calcuation loop after all calculation done.
faster

### DIFF
--- a/src/.stories/index.js
+++ b/src/.stories/index.js
@@ -516,7 +516,7 @@ storiesOf('General | Layout / Grid', module).add('Basic setup', () => {
       <ListWrapper
         component={SortableList}
         axis={'xy'}
-        items={getItems(10, 110)}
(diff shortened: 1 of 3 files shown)


arthurakay:
I'm fairly certain this issue also causes some significant performance issues we're seeing in IE11. I'd love to see this PR merged!

PR #652, opened 2019-12-21
Fixed getScrollingParent and isScrollable when used in iframes

KevinGhadyani-minted:
When using `srcDoc` on an `iframe`, you're pushing HTML in from your parent `window` into the `iframe` directly. Because of this, we need to reference the element's instances of `HTMLElement` and `window` to ensure we're using the right values.

For instance, an element made in an iframe has a different `HTMLElement` class instance than the one in the parent document which this JavaScript was imported.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -240,17 +240,16 @@ export function getLockPixelOffsets({height, width, lockOffset}) {
 }
 
 function isScrollable(el) {
-  const computedStyle = window.getComputedStyle(el);
-  const overflowRegex = /(auto|scroll)/;
-  const properties = ['overflow', 'overflowX', 'overflowY'];
-
-  return properties.find((property) =>
-    overflowRegex.test(computedStyle[property]),
-  );
+  var computedStyle = el.ownerDocument.defaultView.getComputedStyle(el);
+  var overflowRegex = /(auto|scroll)/;
+  var properties = ['overflow', 'overflowX', 'overflowY'];
+  return properties.find(function (property) {
+    return overflowRegex.test(computedStyle[property]);
+  });
 }
 
-export function getScrollingParent(el) {
-  if (!(el instanceof HTMLElement)) {
+function getScrollingParent(el) {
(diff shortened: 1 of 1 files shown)


KevinGhadyani-minted:
Sorry, accidentally removed `export` from `getScrollingParent`. This has been fixed.

KevinGhadyani-minted:
Fixed the issues from code formatting when I brought this in from the Babel'd version I was looking at versus the actual code.

KevinGhadyani-minted:
Hi @clauderic. Is there anything I can do to help you feel comfortable merging? I have a hack in our application right now, but I'd like an official one ;) 👍.

KevinGhadyani-minted:
Can we get this merged? Is there any testing you'd like me to do?

NathanCH:
I have an issue with iFrames so I was curious about this PR. I manually applied your changes but get the following:

    TypeError: el.ownerDocument is null

KevinGhadyani-minted:
I see, if you don't have a valid DOM element, then it wouldn't work. Is that the issue you're running into?

As far as I know, all DOM elements have an `ownerDocument` prop. The `defaultView` prop is window on `document` for that particular element's `window`.

image

Sawtaytoes:
I figured it out why you're seeing `null`.

The issue is that none of your HTML elements has a "scroll" prop on it. The browser is somehow magically adjusting the overflow behavior using `visible` instead of `scroll`. I set the scroll container to `overflow-y: auto` for it to work properly.

Ideally, this code, and the code in the actual library, would be searching for some other method to see if this is scroll container rather than looking for an `overflow` CSS prop.

KevinGhadyani-minted:
@NathanCH Should I close this PR or is it still valid today?

NathanCH:
Hi @KevinGhadyani-minted, I wasn't able to solve my problem and ended up moving to [dnd-kit](https://github.com/clauderic/dnd-kit/). Thanks

KevinGhadyani-minted:
While this is technically still valid, I'm closing this PR because I haven't received any response from the developer.

PR #684, opened 2020-04-08
Fix `getEdgeOffset` from `src/utils.js`

832bb9:
`getEdgeOffset` works buggy. It seems the problem is `node.offsetTop`/`node.offsetLeft` returns distance to closest RELATIVE parent node, but we want to get distance to ACTUAL parent node. Replacing `offsetTop`/`offsetLeft` with `getBoundingClientRect` should fix problem.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -158,11 +158,14 @@ export function getEdgeOffset(node, parent, offset = {left: 0, top: 0}) {
   if (!node) {
     return undefined;
   }
-
-  // Get the actual offsetTop / offsetLeft value, no matter how deep the node is nested
+  
+  const nodeRect = node.getBoundingClientRect();
+  const parentRect = node.parentNode.getBoundingClientRect();
+  
+  // get distance to parent edges, no matter how deep the node is nested
   const nodeOffset = {
-    left: offset.left + node.offsetLeft,
-    top: offset.top + node.offsetTop,
+    left: offset.left + nodeRect.left - parentRect.left,
+    top: offset.top + nodeRect.top - parentRect.top,
   };
 
   if (node.parentNode === parent) {


832bb9:
The problem with `offsetParent` is it can be above `parent` in DOM.

src/utils.js
```diff
@@ -158,11 +158,14 @@ export function getEdgeOffset(node, parent, offset = {left: 0, top: 0}) {
...
+  
+  // get distance to parent edges, no matter how deep the node is nested
   const nodeOffset = {
-    left: offset.left + node.offsetLeft,
-    top: offset.top + node.offsetTop,
+    left: offset.left + nodeRect.left - parentRect.left,
+    top: offset.top + nodeRect.top - parentRect.top,
   };
 
   if (node.parentNode === parent) {
```
skytrooper09, line 171: There is another bug here that has been filed and is also affecting my use of this library, as well.  The use of ```node.parentNode``` with a getContainer function can cause a recursive loop that ends with #document.parentNode, which is ```null``` and results in an ```undefined``` return value for offsetEdge.

Changing this to ```node.offsetParent``` resolves this issue and stops the recursion at the scrolling container.

```suggestion
  if (node.offsetParent=== parent) {
```

skytrooper09:
> Agreed I have the same problem, if there is a parent that is not the offsetParent in the chain it breaks.
> 
> So either getBoundingClientRect() or chase offsetParent rather than parentNode and return the summed values when hitting a null or the parent.

This PR seems to use the getBoundingClientRect() approach, but it does not handle the ```null``` case for the return.  As I mentioned in my comment/suggestion, the value of the offset is irrelevant if it just returns ```undefined``` when it hits the ```null``` case.

src/utils.js
832bb9: As was mentioned in PR description the problem is using `offset(Top|Left|Parent)`.

So:
- If `parent` is correct ancestor for `node` we will found correct distance using `getBoundingClientRect` recursively.
- If `parent` isn't correct ancestor current implementation will throw exception at `node.parentNode.getBoundingClientRect` where `node.parentNode` is null.

So i suggests check this case at first line of function `if (!node || !node.parentNode) { return undefined; }`. Should it fix your problem?

skytrooper09: I think the root of my problem maybe instead comes from the fact that this function returns ```undefined``` at all.  The ```animateNodes()``` function in ```SortableContainer/index.js``` uses the offsetEdge.left/right values directly and does not seem to handle the undefined case.

```js
// from animateNodes()
      const sortingOffset = {
        left:
          this.offsetEdge.left + this.translate.x + containerScrollDelta.left,
        top: this.offsetEdge.top + this.translate.y + containerScrollDelta.top,
      };
```

This only happens when I use my own getContainer element, which is a few levels up in the DOM, and not the SortableContainer element.  Using ```offsetParent``` avoids the recursive loop from getting all the way up to #document and ```null```, but even when it does get that far the offset values are still correct, if they were to be returned.

832bb9: I mean `offsetParent` will be node somewhere between `root` (inclusive) and `node`, so recursive `node.parentNode` will reach it anyway. It is not obvious how to handle If `parent` is not ancestor of `node`.

I have bug just because of `offsetParent` and if it works fine in great amount of cases it is still dangerous to use.

Issue #683, opened 2020-04-06
@babel/runtime should be a peerDependency instead of dependency


anna1go:
If `@babel/runtime` were set to be a `peerDependency` it would allow users of this library to choose which version of `@babel/runtime` to install. Our web app already uses multiple different versions of `@babel/runtime`. After installing `react-sortable-hoc`, we saw an increase of 2 KiB due to `@babel/runtime`. By forcing it to be a "dependency" and not "peerDependency", we end up duplicating code.

PR #685, opened 2020-04-09
Move to babel-runtime to peerDependencies

agoncharova:
**Problem:** `@babel/runtime` set as a `dependency` forces our app to download ~2KB of additional data, despite the fact that we already have a version of that package.

**Solution:** We realized that if we move this package to be a `peerDependency` it will allow to opt out of downloading that specific version of the package if other versions exist; therefore saving ~2KB.

This resolves issue [683](https://github.com/clauderic/react-sortable-hoc/issues/683).

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,11 +48,11 @@
     }
   },
   "dependencies": {
-    "@babel/runtime": "^7.2.0",
     "invariant": "^2.2.4",
     "prop-types": "^15.5.7"
   },
   "peerDependencies": {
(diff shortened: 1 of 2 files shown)


TrySound:
You can remove and install this package again to upgrade its transitive dependencies.

Moving it to peer dependency will make package error prone. If your project update only this package and it uses newer babel runtime you may get breaking change.

And nothing will stop you from this mistake except small warning while installing nobody look at.

PR #691, opened 2020-04-29
Fix shift to next row calculation

### LINKED ISSUE #523: tile position calculation is broken
from #520 

image
When an element of a row is moved to an upper one, get a overflow

the problem is seen only on certain conditions.
here is my code https://codesandbox.io/s/w68jykzw58 showing the bug you can fix it by removing overflowX: "hidden" from the wrapper

### LINKED ISSUE #551: Incorrect item positioning in multiline horizontal list
Hi Guys,
I'm trying to use react-sortable-hoc with multiline horizontal lists. It works perfectly well when items are of the same size, but it totally fails to position items properly if they are different:

Screen recording:
https://i.imgur.com/oYTuiZ0.gifv

Simple demo:
https://codesandbox.io/embed/flamboyant-hawking-vwlbq

What do I do wrong? How to fix it?
Tnx!

### LINKED ISSUE #626: 2 Column container problem
Hi guys, first of all, I want to congratulate you for such a good library.

I've encountered a bug on the library, working with 2 columns `SortableContainer`'s,  it has a weird animation when dragging some components.

To reproduce this problem just go to http://clauderic.github.io/react-sortable-hoc/#/basic-configuration/grid?_k=8abqva and edit the width of the div with the classes `Showcase__style__list Showcase__style__stylizedList Showcase__style__grid`  to `width: 350px`
ezgif com-video-to-gif

### LINKED ISSUE #689: Erratic behaviour in Firefox (mac)
Hi, I've noticed that react-sortable-hoc behaves erratically on firefox (works perfectly on chrome/safari).
The ghosted object often won't drag and the actual drag position (i.e where clicked/mouse-down) is different. Best explained in this video
firefox
To reproduce: go [here](https://css-tricks.com/draggin-and-droppin-in-react/) 

Granted that the demo is a 9months old (different rsh version?)... but the behaviour is similar to what I'm seeing in my code.

Thanks :-)

alexcheuk:
This PR fixes the issue that is happening in this demo (http://clauderic.github.io/react-sortable-hoc/#/basic-configuration/grid?_k=8abqva)

and demonstrated in this issue #626 
image

The calculation that was made to determine if a node has moved past the right bounds is using `offset.width` which is only half of the width of the node. This means, if there are more than half of the width of empty space on the right of the container, the node will shift out of the Grid like the image above.

resolves #626, resolves #551, resolves #523, resolves #689

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -727,7 +727,7 @@ export default function sortableContainer(
               translate.x = this.width + this.marginOffset.x;
               if (
                 edgeOffset.left + translate.x >
-                this.containerBoundingRect.width - offset.width
+                this.containerBoundingRect.width - offset.width * 2
               ) {
                 // If it moves passed the right bounds, then animate it to the first position of the next row.
                 // We just use the offset of the next node to calculate where to move, because that node's original position


23chrischen:
+1 I'd love to see this merged, this is causing me some headache

alexcheuk:
> +1 I'd love to see this merged, this is causing me some headache

I temporarily published the fix at `@alexcheuk/react-sortable-hoc` if anyone wants to test if it fixes their issues.

Gordon-Feng:
`offset.width * 2` does not work for me, but `offset.width * 4` is work.

Issue #648, opened 2019-12-17
disabled option still allows an element to move


stoneshower:
I see  with `disabled` option the element is not draggable anymore but still moves when I drag another element over it. How can I make it "fixed" position and not let it change the order.
In my case, I need to set the first item fixed, so it should not be draggable and unsortable.

stoneshower:
I tried to cancel the sort action by passing condition on `onSortMove` and `return` but did not work either.

patrikw1:
> I tried to cancel the sort action by passing condition on `onSortMove` and `return` but did not work either. I need to a function like `shouldCancelEnd`
> 
> After a few hours investigation, I managed to add a function like below,
> 
> ```
> onSortEnd={({newIndex}) => {
>          if (newIndex === 0) { // the first element is "unsortable"
>              return;
>          }
>    }
> }
> ```
> 
> But ideally I don't want the element move even though it goes back to the original position after mouse unholding

I had a same issue, when first element should be "unsortable". In this case my workaround was

```
<SortableItem collection={index === 0 ? 1 : 0}/>
```

This only applies if "unsortable" item is a first item.

keshavashiya:
> > I tried to cancel the sort action by passing condition on `onSortMove` and `return` but did not work either. I need to a function like `shouldCancelEnd`
> > After a few hours investigation, I managed to add a function like below,
> > ```
> > onSortEnd={({newIndex}) => {
> >          if (newIndex === 0) { // the first element is "unsortable"
> >              return;
> >          }
> >    }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But ideally I don't want the element move even though it goes back to the original position after mouse unholding
> 
> I had a same issue, when first element should be "unsortable". In this case my workaround was
> 
> ```
> <SortableItem collection={index === 0 ? 1 : 0}/>
> ```
> 
> This only applies if "unsortable" item is a first item.

```
<SortableItem collection={index === 0 ? 1 : 0}/>
```
This will work for both starting and ending elements just need to pass `collection={1}`.

PR #693, opened 2020-05-05
Locked items

cblagg:
This relates to the issue at: https://github.com/clauderic/react-sortable-hoc/issues/648

### What does this do?
It implements the ability for configuring "locked items". Locked items are items that are not draggable, and also do not move when an item is dragged over them.

I implemented this as `lockedItems` distinct from disabled items - given potential for wanting elements that are fixed in their position, but not necessarily in a "disabled" visual state. To have both disabled and locked state items, one would pass the distinct disabled and locked props.

### How did I test this?
I implemented stories for horizontal, vertical, and grid layouts, and dragged items around in different scenarios and velocities. I was unable to find an existing testing setup to write additional tests - please let me know if there is something I overlooked in this regard.

A video of the interaction and stories this PR introduces can be found at:
https://drive.google.com/file/d/1sS2vh161ai9IHaIRDEshsAyicWkdFG3h/view?usp=sharing

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -37,6 +37,11 @@ import {
   defaultKeyCodes,
 } from './props';
 
+const DIRECTION = {
+  FORWARD: 'forward',
+  BACKWARD: 'backward',
+};
+
 export default function sortableContainer(
   WrappedComponent,
   config = {withRef: false},
@@ -143,9 +148,9 @@ export default function sortableContainer(
         !this.state.sorting
       ) {
         const {useDragHandle} = this.props;
-        const {index, collection, disabled} = node.sortableInfo;
+        const {index, collection, disabled, locked} = node.sortableInfo;
 
-        if (disabled) {
+        if (disabled || locked) {
           return;
         }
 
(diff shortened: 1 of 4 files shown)


src/SortableContainer/index.js
```diff
@@ -636,9 +641,43 @@ export default function sortableContainer(
       const prevIndex = this.newIndex;
       this.newIndex = null;
 
+      // Returns whether a node location is a valid location for another node
+      // to be moved into
+      const isAllowedToMoveToNode = ({node}) => !node.sortableInfo.locked;
```
cblagg, line 646: If this becomes:

```
({node}) => !node.sortableInfo.locked && !node.sortableInfo.disabled
```

This performs the same positional locking for disabled items that is requested in: https://github.com/clauderic/react-sortable-hoc/issues/648

clauderic:
Hey there, thanks for your contribution. I'm just wondering whether this feature is really needed. Right now you should be able to achieve the same behaviour by not wrapping in `SortableElement` the elements you wish to behave as "locked"

cblagg:
Unfortunately I don't think the same behavior can be reproduced by only conditionally wrapping (or not wrapping) items with `SortableElement`. 

When I do that, while it means those elements don't move when an item is dragged over it, the items on either side of those do not complete animations to bypass the locked elements. Instead animating to just sit on top of a locked item. 

But perhaps I'm missing some configuration that would change this? Though from what I gathered from reading `animateNodes` the existing calculation of transitions doesn't have hooks that would allow for maneuvering this.

A video of what I'm seeing when only conditionally wrapping is: https://drive.google.com/file/d/1SOlSKB5sIl91CMMdGBvNsPcqeR03oz5P/view?usp=sharing the difference being visible in comparison to the video in my opening description.

scottastrophic:
@clauderic Hi there, longtime fan of your buttery smooth transitions. FWIW I hope you give this solution some consideration... On my project (an automated-email composer with certain fixed sections around which other, customizable sections can be dragged), we ended up forking the library to achieve a similar effect. I concede the disabled-vs-locked differences are not stark; if starting from scratch I would probably vote for a unified implementation where the affected items were neither grabbable nor serving as displacement targets for unaffected items, but this solution has the benefit of backwards compatibility. It really completes the effect to have the locked items as stable, unlandable gaps within the list. Thanks for considering!

cblagg:
I took another look at this, given the suggestion from @clauderic regarding just conditionally wrapping an item with `SortableElement`.

As shared in the video in my response, currently on master, conditionally not wrapping an item with SortableElement, leads to other draggable items being positioned on top of them when they are displaced during a drag event.

At: https://github.com/clauderic/react-sortable-hoc/compare/master...cblagg:cblagg/correct-positioning?expand=1

I've taken the positioning logic updates from this PR, but removed the notion of `locked` making it into the internals. Instead, in the stories `lockedItems` is now just used to conditionally wrap items for demo purposes.

@clauderic if you have no/fewer objections to this updated approach - I'm happy to create a PR out of it (or merge that changeset into this branch).

timofarei:
Hi there, this positioning logic would help me a lot in one of my lists.

I tried to conditionally wrap my list items with `SortableElement` but end up having the same animation issue described in https://github.com/clauderic/react-sortable-hoc/pull/693#issuecomment-624265954 by @cblagg.

Is this positioning logic available somewhere or was this proposal discarded in the end?

ps: Thanks a lot for the efforts @cblagg

PR #749, opened 2020-10-19
added conditional prop to disable animation when dragging

yuri-scarbaci-lenio:
Same as https://github.com/clauderic/react-sortable-hoc/pull/482 but with latest code-base from master

As per the conversation in the previous PR disabling animation is especially useful if the the final user wants to handle it's own animation.

Even more useful when the ghost item is kept in the list.

This is strictly required for anyone that wants to use `react-sortable-hoc` to implement a custom `tree-view` component, where the animation between different nest levels results more problematic than useful.

In our specific case we want to only add borders between new positions instead of actually moving the objects.

While this may not be the most common use-case, I don't see a reason why not providing a flag to disable the translate, this will never require any more complexity, it will always be a simple toggle that kept everything else as is except for the on drag-over animation.

Probably the code could be improved to be debounced a little, to avoid invoking the animateNode() function so many times when the mouse didn't move at all (start drag, don't move mouse, check how many times the animateNode function is invoked...) or if not debounced, at the very least memoized?


jmangel:
Will this PR/feature ever be merged?

PR #768, opened 2021-01-25
Add react v17 to the peer deps.

pleunv:
Hiya!

Not sure if you're still taking PRs, but just in case a quick one in order to add React v17 to the peer deps. Fwiw, I've tested it on v17 and found no issues.


nelsonmestevao (approved):
:pray:

jonioni:
Should the PR be merged?

Develliot:
@clauderic this library is awesome and we and would love to use it with react 17, could you help us?  🙏

Issue #577, opened 2019-07-25
useWindowAsScrollContainer should not crawl to root node


knarkunas:
There is an issue with grid and `useWindowAsScrollContainer`. It should not be crawling up to first `overflow: scroll|auto` DOM element. In result we get something like this using grid layout:
![image_sort_bug_known](https://user-images.githubusercontent.com/37063113/61844495-511b6780-aea8-11e9-86ad-3919d6567684.gif)
![image_sort_bug_example](https://user-images.githubusercontent.com/37063113/61844502-5678b200-aea8-11e9-9e66-959ff52e0004.gif)

Or we could get a `prop` of some sort for grid layout when scroll is not used at all?
Like `isScrollable` and then this code:
```
this.scrollContainer = useWindowAsScrollContainer
          ? this.document.scrollingElement || this.document.documentElement
          : getScrollingParent(this.container) || this.container;
```

becomes

```
if (isScrollable) {
     this.scrollContainer = useWindowAsScrollContainer
               ? this.document.scrollingElement || this.document.documentElement
               : getScrollingParent(this.container) || this.container;
} else {
      this.scrollContainer = this.container;
}

```

Currently `animateNodes` inside `SortableContainer.js` does transformation and calculation ends up `this.containerBoundingRect` having to be root element (most websites has overflow scroll somewhere, right?).
`edgeOffset.left + translate.x >  this.containerBoundingRect.width - offset.width`

Current workaround, if you want to have non scrollable, simple grid box layout and draggable list is to add `<div style={{ overflow: 'auto' }}>` around your `SortableContainer`.

Example to reproduce:
https://codesandbox.io/s/serene-dawn-rwpj0

knarkunas:
Possible other issues related to this one:
https://github.com/clauderic/react-sortable-hoc/issues/551
https://github.com/clauderic/react-sortable-hoc/issues/523
https://github.com/clauderic/react-sortable-hoc/issues/520

zalevsk1y:
It began with version 1.8.0. Before this, in version 1.7.1, the parent container was simply taken for the calculation.

`const containerBoundingRect = this.container.getBoundingClientRect();`

Mathieu-DA:
Thanks knarkunas! Your workaround saved my day. It's the 'overflow: auto' that did the trick.
I didn't have to set useWindowAsScrollContainer to false.

PR #806, opened 2021-07-16
Fix bug with window scroll container

akhmadullin:
### Problem

There were some issues about incorrect sorting behavior in grid list with enabled ``useWindowAsScrollContainer``, for example, https://github.com/clauderic/react-sortable-hoc/issues/577.

When prop ``useWindowAsScrollContainer`` is set to ``true``, sorting in grid list does not work properly:

https://user-images.githubusercontent.com/37641303/125896642-c22e9532-d075-449c-bccc-ce96d08ce47c.mp4

### Solution

Such behavior is caused because of using root ``html`` element as scrolling container. Scrolling container's width is used for counting of new coordinates of list item during sorting. ``html`` element has width much more, than grid list container element, new coordinates counts wrong, that's why we can see incorrect sorting behavior.

This PR fixes issue with incorrect sorting by using width of grid container instead of using width of ``html`` element, where it's needed.

The result of this fix:

https://user-images.githubusercontent.com/37641303/125897590-8ffa7933-2b37-4bc8-a714-4dae347fbae8.mp4

### DIFF
--- a/src/.stories/Storybook.scss
+++ b/src/.stories/Storybook.scss
@@ -11,6 +11,10 @@ $focusedOutlineColor: #4c9ffe;
   align-items: center;
 }
 
+.rootAutoHeight {
+  height: auto;
+}
+
 // Base styles
 .list {
   width: 400px;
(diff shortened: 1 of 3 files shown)


lavish0000:
OK, Got it. Thanks @akhmadullin for the reply.

PR #833, opened 2022-03-05
fix: Fix dragging on iOS when zooming

IArny:
This commit is fixing a bug when user of iOS zooming screen with sortable elements. When two sortable items are touched then ```handlePress``` function fires two times and after releasing item one of them runs away from screen.

https://user-images.githubusercontent.com/42856876/156889617-dbb30691-3ce9-4ae3-826a-cb9d6851b3bb.mp4


IArny:
@clauderic review pull request please, really need this fix

PR #844, opened 2022-06-02
Support React v. 18

patrykboch:
Fixes https://github.com/clauderic/react-sortable-hoc/issues/838

PR #846, opened 2022-07-10
Allow installation with react 18

alexander-schranz:
There seems not lot of changes in react 18: https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html so I think this should be possible without any efforts.


alexander-schranz:
closing in favor of #844

PR #849, opened 2022-07-31
fix:Fix sortableContainer distance inaccurate

1464633490:
Fix the distance attribute. Change the offset sum of X and Y axes to the pixel value of a straight line

### DIFF
--- a/src/SortableContainer/index.js
+++ b/src/SortableContainer/index.js
@@ -15,6 +15,7 @@ import {
   getElementMargin,
   getLockPixelOffsets,
   getPosition,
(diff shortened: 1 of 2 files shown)


1464633490:
Using the distance attribute of the container, according to the description document and usage, it should be the linear pixel value of the offset position and the origin, rather than the sum of the offsets of the X and Y axes. I hope the author can see and merge this PR, otherwise the description of distance is ambiguous.

vaynevayne:
Also encountered this problem, now this library, do not repair bug, nor allow others to pr tmd

PR #869, opened 2023-02-15
Update types in types/index.d.ts

hpello:
# Description

The current types trigger an error:
```
node_modules/react-sortable-hoc/types/index.d.ts:108:11 - error TS2694: Namespace 'React' has no exported member 'SFC'.

108   | React.SFC<P>
              ~~~
```

Indeed, `React.StatelessComponent` has been removed from the react types (see https://stackoverflow.com/questions/44375759/how-should-i-declare-a-stateless-functional-component-with-typescript-in-react)

This PR fixes this error.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -105,7 +105,7 @@ export type WrappedComponentFactory<P> = (props: P) => JSX.Element;
 
 export type WrappedComponent<P> =
   | React.ComponentClass<P>
-  | React.SFC<P>
+  | React.FunctionComponent<P>
   | WrappedComponentFactory<P>;
 
 export function SortableContainer<P>(


hpello:
Hello @clauderic! Could you take a look at this PR? 🙂 
We need it to keep using `react-sortable-hoc` in our TS codebase.

ErikBrendel:
I also just stumbled across this problem.

(Explicitly passing the prop type to `SortableElement<P>` and `SortableContainer<P>` helps as a workaround)

Having this update would still be nice though!

hpello:
Hi @clauderic, any chance you can have a look at this PR? 🙂 

- For users with `skipLibCheck: false`, it will allow using this library
- For users with `skipLibCheck: true`, it will catch misuses of the library! Else component props are typed as `any` 😞 

🙏

hpello:
closing for housekeeping
feel free to reopen

